PR #15, opened 2015-01-23
Make SearchResult objects JSON.stringify'able

grncdr:


index.js
```diff
@@ -139,21 +139,24 @@ var Field = redefine.Class({
...
 var SearchResult = redefine.Class({
-  constructor: function SearchResult() {},
+  constructor: function SearchResult(ItemType, object) {
+    this.responseData = object;
+    this.total = object.total;
+    this.limit = object.limit;
+    this.skip = object.skip;
+    walkMutate(_.cloneDeep(object), isParseableResource, parseResource);
+    resolveResponse(object).forEach(function (item, i) {
+      this[i] = item;
```
madtrick: why are you assigning in the SearchResult instance every item to a property named after its index?

grncdr: It's intended to be backward compatible with the current API (which returns an augmented array).

grncdr: Although actually that makes no sense at all so maybe I will try something else.

index.js
```diff
@@ -138,24 +138,30 @@ var Field = redefine.Class({
...
-          enumerable: false
-        }
-      );
-    }
-  }
+SearchResult.prototype.toJSON = function () {
+  return this.responseData;
+};
+
+Object.keys(Array.prototype).filter(function (key) {
```
sdepold: Why are we doing this?

test/integration/content-type-test.js
```diff
@@ -16,7 +16,7 @@ buster.testCase('Content Type', {
       'order': 'name',
       'sys.id[in]': 'cat,dog,human'
     }).then(function(contentTypes) {
-      assert.match(contentTypes, [
+      assert.match(contentTypes.items, [
```
sdepold: Why did this change? Isn't that a BC breaking change?

madtrick:
@grncdr what's the current state of the PR?

test/integration/content-type-test.js
grncdr: yes, as commented in the commit message, this would break BC. The old API is not tenable.

grncdr:
@madtrick - test failures need fixing and then we need to bump major version.

grncdr:
This could be re-opened/worked on by anybody in the future, but I don't have the time to finish it now.

PR #18, opened 2015-03-02
Remove unused itemType parameter

trodrigues:
I checked old commits and this is a leftover that is not used anymore


sdepold:
LGTM

PR #19, opened 2015-03-22
Use axios instead of questor

trodrigues:
The main reasons for this are:
- Use a more widespread and maintained library than questor
- Fix an issue that seems to have been introduced with a recent update of questor, which introduces issues with CORS requests without a seemingly easy fix.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict';
 
 var _ = require('lodash');
-var questor = require('questor');
+var axios = require('axios');
 var redefine = require('redefine');
 var resolveResponse = require('contentful-resolve-response');
 var querystring = require('querystring');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -19,72 +19,78 @@ var Client = redefine.Class({
 
   request: function(path, options) {
     if (!options) options = {};
-    if (!options.headers) options.headers = {};
```
sdepold: It seems to me that the user cannot provide custom headers anymore. While I cannot see the usecase for that, we are still breaking BC. Is that on purpose?

sdepold:
I think the rest looks good.

sdepold:
Released as 1.0.0

PR #20, opened 2015-03-23
Add support for the sync endpoint

trodrigues:


test/integration/sync-test.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+buster.testCase('=> Sync', {
```
sdepold: Please remove the focus rocket

index.js
```diff
@@ -90,6 +90,33 @@ var Client = redefine.Class({
...
+                   return {
+                     items: sync.items,
+                     nextSyncToken: sync.nextSyncToken
+                   };
+                 }
+               });
+  },
+
+  sync: function (object, callback) {
+    var query = Query.parse(object);
```
sdepold: Shouldn't we do any kind of sanity check here?

index.js
```diff
@@ -90,6 +90,33 @@ var Client = redefine.Class({
...
+    var items = [];
```
sdepold: What do we need this variable for ?

test/integration/sync-test.js
```diff
@@ -0,0 +1,14 @@
...
+  prepare: function() {
+    this.timeout = 90e3;
+  },
+
+  'can be read': function() {
```
sdepold: Could we add a test that does the initial sync, adds an entry and continuous the sync with the nextSyncUrl, which should then return the just created entry?

sdepold:
I added a change log in master. Please rebase and add your notes there :)

test/integration/sync-test.js
trodrigues: oops!

index.js
trodrigues: good point

index.js
trodrigues: Ah good catch. I wrote an initial version of the code that was very different, then refactored stuff into the Sync class and this is somehow a leftover.

test/integration/sync-test.js
trodrigues: Yeah, I thought about a way of testing the next sync but there's no good way of doing that.

sdepold:
LGTM

PR #23, opened 2015-03-31
Convert to ES6

grncdr:
There's (almost) no logic changes here, just introducing babel for compiling classes.


sdepold:
Please point the package.json to a file that falls back to the babel register mechanism: https://github.com/contentful/contentful-node-dev-tools/blob/master/index.js

grncdr:
Thanks for pointing that out, it also made me realize the npm packaging would have been missing `index.es5.js`. That's sorted out now, and as a bonus I added a `"watch"` script so that `npm run watch` will work for development.

index.es6.js
```diff
@@ -0,0 +1,305 @@
+'use strict';
+
+var _ = require('lodash');
+var axios = require('axios');
+var resolveResponse = require('contentful-resolve-response');
+var querystring = require('querystring');
```
geigerzaehler: Since we are going ES6 here, we might want to use the [`import`](https://babeljs.io/docs/learn-es6/#modules) statement.

index.es6.js
```diff
@@ -0,0 +1,305 @@
...
+    });
+  }
+}
+
+class ContentType {
+  static parse (object) {
+    return _.extend(new ContentType(), {
+      sys: Sys.parse(object.sys),
+      fields: object.fields.map(Field.parse),
+    }, _.pick(object, 'name', 'displayField'));
```
geigerzaehler: [Destructuring](https://babeljs.io/docs/learn-es6/#modules) might make this more readable
~~~js
static parse ({name, sys, fields, displayField}) {
  return _.extend(new ContentType(), {
    name: name,
    sys: Sys.parse(sys),
    fields: fields.map(Field.parse),
    displayField: displayField
  })
}
~~~

geigerzaehler:
Let’s run this through JSCS while we’re at it. Otherwise: :+1:

grncdr:
This is good to go on my side.

geigerzaehler:
Why was this closed?

grncdr:
It was open for 10 days, feel free to re-open at your convenience, I'm just clearing out my GH queue.

geigerzaehler:
Last call for comments. Otherwise I will merge this tomorrow.

package.json
```diff
@@ -1,30 +1,38 @@
...
     "saucie": "^0.1.0",
     "testem": "~0.6.2",
-    "uglify-js": "^2.4.19"
+    "uglifyify": "^3.0.1"
   },
   "scripts": {
+    "jscs": "jscs --esnext index.es6.js",
+    "bower-build": "browserify -s contentful --debug -t [ babelify --sourceMapRelative . ] -t uglifyify index.es6.js | exorcist dist/contentful.min.js.map > dist/contentful.min.js",
+    "prepublish": "babel index.es6.js -s -o index.es5.js",
+    "watch": "babel index.es6.js -s -w -o index.es5.js",
```
johanneswuerbach: Why do we have `index.es5.js`, `index.es6.js` and `dist/contentful.min.js` now? Shouldn't `index.es5.js` stay `dist/contentful.js`?

sdepold: yeah I'm confused too. also wouldn't call it `index.es6.js` but just `index.js`. In the end it's "normal" js.

grncdr:
Sorry for the confusion, was signed off by @geigerzaehler and @trodrigues. Will reply to last comment above.

package.json
grncdr: The `index.es5.js` is what is used on node, while the `dist/contentful.js` is minified and browserified for browser clients.

PR #24, opened 2015-03-31
Add getters for Asset properties

grncdr:
Depends on #23


sdepold:
broken btw

grncdr:
Yes, it's still based on a broken commit from #23 . will rebase after that is merged

PR #25, opened 2015-04-02
Update dist files

### LINKED ISSUE #22: Problem with snyc
Hi. I've pulled the latest contentful.js and am trying to test the synchronization and it gives me an error : 

Uncaught TypeError: undefined is not a function

at 

// Get all data in a space
        client.sync({initial: true})
            .then(function(data){
                syncToken = data.nextSyncToken;
            });

Any idea what is going on? I've copied your example exactly. The only thing I've changed is the space id and access token.

sdepold:
Fixes #22


trodrigues:
Ok, then this looks good I'd say.

PR #33, opened 2015-08-14
Add API documentation

grncdr:
This is pretty much a copy-paste from the docs that used to exist on contentful.com


trodrigues:
Can you also add this to contentful-management.js ?

PR #53, opened 2016-01-06
Added `publicEntries` call. Fixed linting errors.

dberesford:


trodrigues:
Hi @dberesford 

this endpoint is redundant for contentful.js, as all the content returned by contentful.js is already published, as it comes from the Delivery API which only Delivers published content.

The `/public/entries` endpoint only applies to the Management API , where if you request something from `/entries` you will get entries which are both published and unpublished.

dberesford:
Hi Tiago,

thanks for the quick response - we are using the management api and need to
filter published from unblished. What do you recommend here?

trodrigues:
Well, first of all, the management API has support for that endpoint with the `getPublishedEntries` call: https://github.com/contentful/contentful-management.js#spacegetpublishedentriesquery---entrycollectionpromise

Second, even if you get all the entries with the `getEntries` call, you should have a property under `sys` called `publishedVersion`. This property will only be there if the entry was ever published.

You will also have a `version` property. The version property is always increased on the act of publishing, so an entry that was published at version 3, will have a `publishedVersion` of 3, and a `version` of 4. Essentially, if `publishedVersion +1 > version`, that means the entry was published, but someone has made additional changes to it afterwards without publishing those changes.

I'll make note of this and improve the documentation or perhaps also add a method to verify this.

Let me know if that helps.

PR #54, opened 2016-01-07
update axios version and use https-proxy-agent

forrestbthomas:
This should enable developers to use this NPM module from behind a proxy.


trodrigues:
Hello @forrestbthomas,

first of all, thanks a lot for your contribution.

However, this is something we'd like to keep outside of the scope of this library, as adding some of these extra dependencies is relevant only for node and makes the browser version heavier.

I've updated axios separately, and added the possibility to pass down the agent in the same way axios allows, so this should also solve your problem.

It's available on version 2.1.0.

forrestbthomas:
Makes sense. Thanks! :+1:

PR #73, opened 2016-05-06
Import individual lodash modules to make the bundle smaller

fson:
This cuts down the size of the minified and gzipped browser bundle by ~30% and helps me reduce the code size when using Contentful from npm in my webpack build.

I left the imports in test modules unchanged, because they won't be included in the final build.

### DIFF
--- a/lib/create-request-config.js
+++ b/lib/create-request-config.js
@@ -1,4 +1,4 @@
-import {cloneDeep} from 'lodash/lang'
+import cloneDeep from 'lodash/cloneDeep'
 
 /**
  * Creates request parameters configuration by parsing an existing query object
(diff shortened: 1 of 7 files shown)


trodrigues:
Hi there, thanks for sending this.

That's really good to know, that it actually cuts down on the bundle size. When I started rewriting this library in ES6 using destructuring on the lodash modules I think they were not yet available as individual methods (pre lodash 4, but I could be wrong).

I then made some experiments with rollup and some tree shaking solutions to remove the dead code but had some issues so I held off on using that until later on.

I had learned the new versions of lodash were importable as individual methods but never really bothered making the changes.

I'm just going to hold off on merging this for a couple of days as I'm about to merge a branch where I'm factoring some code out of this library into a separate module.

fson:
@trodrigues Yup, that makes sense. I think individual modules were added in lodash v4. Let me know, if you want the test cases converted to the same style for consistency. They don't have an effect on the bundle size, however.

trodrigues:
No, that's fine. Thank you :)

Issue #59, opened 2016-02-24
Fix webpack usage


trodrigues:
The npm package only bundles the built files under dist, but webpack doesn't like the fact that there is a require for a `lib` dir that doesn't exist.

This can be fixed by including `lib` but there should be a better way to make webpack not worry about this.

zamiang:
Hey @trodrigues I'm having this problem now. How did you end up resolving this issue?

trodrigues:
Hi there,

yes, I did fix this at the time, can't remember on which specific version for sure. Which version are you using at the moment?

zamiang:
Running the latest contentful (3.3.6) in a webpack + babel project. 

It appears that even though the lib require is in a error handler that is never called, webpack still tries to resolve the require. I forked and removed these lines and all is well. https://github.com/contentful/contentful.js/blob/master/index.js#L7-L9

trodrigues:
Ah yes, that part of the code is there for development mode.

The thing is, webpack should honor the "browser" key on package.json and use the browser.js file instead of index.js.

Could this be an issue with your webpack configuration?

If you'd like to check an example of an app we have that is currently using this, you can have a look at https://github.com/contentful/discovery-app-react/blob/master/webpack.config.js

Koded:
I think this is still a problem (only for serverside nodejs). I'm just starting to use the contentful library and it's taking too long to make it work with webpack and node.

The initial problem was the one outlined in this ticket:

```
ERROR in ./~/contentful/index.js
Module not found: Error: Can't resolve './lib/contentful' in '/Users/jon/Workbench/Joel-Trello/node_modules/contentful'
```

The quick fix for this is the solution that @zamiang outlined in https://github.com/contentful/contentful.js/blob/master/index.js#L7-L9

You can fix it with webpack, but the solution doesn't reflect the changes from Webpack 1 to version 2. Specially, to tell webpage to use `./browser.js` rather than `main.js` you need to specify:

```
resolve: {
  ...
  mainFields: ["browser", "main"]
}
```

so the `webpack.config.js` changes outlined by @trodrigues aren't quite right for webpack v2.

This is a bit funny anyway as you are telling node to use `./browser.js` in `package.json`:

```
"node": "./browser.js"
```

When you get this working, you then need to get webpack to rewrite the reference to the browser version of axios (something like - this isn't working for me yet):

```
    plugins: [
[...]

PR #76, opened 2016-06-03
Support building for a server-side target #59

zamiang:
After our discussion in #59, I did a little digging and found what was wrong (I think?)

Unlike in your example app which is building for browsers, I am using webpack to build for a node target. Since there was no node option in the `package.json`, it would fall back to `index.js` and error trying to load `lib/contentful`.

This PR adds a node option to the package.json to support server-side builds such as this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "semantic-release": "semantic-release pre && npm publish && semantic-release post"
   },
   "browser": "./browser.js",
+  "node": "./browser.js",
   "files": [
     "browser.js",
     "index.js",


Khaledgarbaya:
Hi,

I think the problem here is your webpack configuration.

Webpack will load modules from the node_modules folder and bundle them in. While this is fine for frontend code, backend modules aren't prepared for this.

It would be nice if you share your webpack configuration or try to add this to your webpack.config.js

```javascript
var fs = require('fs');

var nodeModules = {};
fs.readdirSync('node_modules')
  .filter(function(x) {
    return ['.bin'].indexOf(x) === -1;
  })
  .forEach(function(mod) {
    nodeModules[mod] = 'commonjs ' + mod;
  });
```
ps: I just tried contentfuljs in a small express app using webpack and I had no problem

Please let us know if this helped.
Best,
Khaled

zamiang:
Hey @Khaledgarbaya 

Here is a webpack config where it fails for me. I believe the key here is the `target: "node"` declaration but perhaps may have something to do with babel-loader.

```
var commonLoaders = [
  {
    /*
     * TC39 categories proposals for babel in 4 stages
     * Read more http://babeljs.io/docs/usage/experimental/
     */
    test: /\.js$|\.jsx$/,
    loader: 'babel-loader',
    include: path.join(__dirname, '..', 'app'),
    exclude: path.join(__dirname, '..', 'node_modules')
  },
  { test: /\.json$/, loader: "json-loader" },
  { test: /\.css$/,
    loader: ExtractTextPlugin.extract('style-loader', 'css-loader?module!postcss-loader')
  }
];

module.exports = {
    name: "server-side rendering",
    context: path.join(__dirname, "..", "app"),
    entry: {
      server: "./server"
    },
    target: "node",
    output: {
      path: assetsPath,
      filename: "server.js",
      publicPath: publicPath,
      libraryTarget: "commonjs2"
    },
    module: {
      loaders: commonLoaders
    },
    resolve: {
      root: [path.join(__dirname, '..', 'app')],
      extensions: ['', '.js', '.jsx', '.css']
    },
    plugins: [
      new webpack.optimize.OccurenceOrderPlugin(),
      new ExtractTextPlugin("styles/main.css"),
      new webpack.optimize.UglifyJsPlugin({
        compressor: {
          warnings: false
        }
      }),
      new webpack.DefinePlugin({
        __DEVCLIENT__: false,
        __DEVSERVER__: false
      }),
      new InlineEnviromentVariablesPlugin({ NODE_ENV: 'production' })
    ],
    postcss: postCSSConfig
  }
```

Khaledgarbaya:
Hi,

I think it is related to the modules loading in my config I am using externals 
So I have this. 
```javascript
var nodeModules = fs.readdirSync('node_modules')
  .filter(function(x) {
    return ['.bin'].indexOf(x) === -1;
  });
//......
var backendConfig = config({
  entry: [
    './src/main.js'
  ],
  target: 'node',
  output: {
    path: path.join(__dirname, 'build'),
    filename: 'backend.js'
  },
  node: {
    __dirname: true,
    __filename: true
  },
  externals: [
    function(context, request, callback) {
      var pathStart = request.split('/')[0];
      if (nodeModules.indexOf(pathStart) >= 0) {
        return callback(null, "commonjs " + request);
      };
      callback();
    }
  ],
//...... rest of the config

```
Can you try it and tell me if it is working.
Best,
Khaled

zamiang:
Interesting. That does fix the issue for me. 

I really don't think adding that code is an acceptable solution for me. I would rather not introduce sophisticated logic into externals -- it is intended for things like `externals: { jquery: "jQuery" }`. Hopefully this test provides some insight into the root of the problem leading to a more sustainable fix.

FWIW, matching the node target to an appropriate file in package.json seems like a more appropriate solution. I haven't (knowingly!) encountered a library where the default main js file in package.json is not intended for primary use. Given that webpack supports many targets beyond 'node' and 'browser', maybe there are other patterns for setting up files for various targets that are more appropriate.

Khaledgarbaya:
Thanks for all the feedback, I will definitely look into that more with the team.
Have a great weekend :)

trodrigues:
Hello @zamiang 

I was having a better look at this (I was away for the last week so couldn't look into this immediately), as I setup the current build, and I was investigating this a bit better and doing a few more tests.

Your solution does indeed seem to fix the issue and from what I can tell so far, it looks like the only thing we could really do from our side that could help out with people using targets in webpack other than browser.

However, this is one of those package.json fields (like `browser`) which are somehow "non standard", so what I'd like to understand is what it means and exactly and where it comes from, as I can't find any source of documentation for this, either on the webpack docs or anywhere else.

What I'm trying to understand is: is the `node` field in package.json a webpack thing, or is it maybe just a coincidence that this works? Also, if we'd add this, what impact could this possibly have on other build tools and/or JS environments?

I'll do some more investigation on this, but if you could provide a source of information for where this field comes from, that would be very helpful, as otherwise I'd feel a bit weary of adding something when we can't truly understand the full impact it would have.

zamiang:
Quick update, found some additional info about webpack targets. They are documented here: https://webpack.github.io/docs/configuration.html#target

I am still in a scenario where I do need to use the `node` target functionality and am unable to use the default `web` target.

Alternatively, if this is not a change you are interested in, I would love some guidance on how I might use a fork the project. Today's attempts have not been successful due to changes that happen when the module is packaged and published to npm vs how files appear in the repo.

Khaledgarbaya:
Hi @zamiang ,
Sorry I was away for a little bit. Let me take a look and get back to you asap.
Best,
Khaled

Khaledgarbaya:
Hi @zamiang ,
I will merge the pull request as it turns to be the only solution to your issue and I will keep an eye on the lib hopefully there is no side effects. Thanks for the contribution
Best,
Khaled

PR #87, opened 2016-08-29
feat(contentful): Added getAllEntries

Munhoziii:
Added support to cross result-page references before the entire object graph is built

### DIFF
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -148,6 +148,38 @@ export default function createContentfulApi ({
   }
 
   /**
+   * Gets a collection of Entries even if they span over a page of results
+   * @memberof ContentfulClientAPI
+   * @param  {Object=} query - Object with search parameters. Check the <a href="https://www.contentful.com/developers/docs/javascript/tutorials/using-js-cda-sdk/#retrieving-entries-with-search-parameters">JS SDK tutorial</a> and the <a href="https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters">REST API reference</a> for more details.
+   * @param  {boolean=} query.resolveLinks - When true, links to other Entries or Assets are resolved. Default: true.
+   * @return {Promise<Entities.EntryCollection>} Promise for a collection of Entries
+   * @example
+   * client.getAllEntries({content_type: 'contentTypeId'})
+   * .then(entries => console.log(entries.items))
+   */
+  function getAllEntries (query = {}) {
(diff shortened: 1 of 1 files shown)


Munhoziii:
The `getEntries` function only fetches a single page of results and builds out the object graph to return it to the caller. There is also a need to have the entire object graph built so that references from multiple pages are also resolved to single instances of those objects, which doesn't seem to be addressed by this lib so far. We were in written communication with Contentful staff about this some months ago and currently have a support ticket related to the need.

Khaledgarbaya:
Hi @Munhoziii,

Thank you for the pull request but I don't see the need to have such method in the SDK, we'd prefer it to be on Application side as this can be replicated easily using the already existing getEntries with combination of skip and limit. I can see a problem for example when users have millions of entries, it is not a good idea to request all of them at once.
I understand that this will involve more code in your app but the end result will be better and you will have more control over it as dealing with paginated data performs better when dealing with huge data sets.

Best,
Khaled

Munhoziii:
Hey @Khaledgarbaya ,
Thank you for the response and for looking into it. I do agree this would go really bad with millions of entries but we have seen use cases as described in #88 and we have described them to the support team for over 6 months now. We basically just need a good way around them - and working with the dereferenced graph to mitigate them isn't easy, and so isn't rebuilding the library on our applications to deal with those cases.
Do you see better ways of getting the same results?
I can see how we could make this better adding a limit to the pages fetched at once or letting the user parameterize them but at the end of the day for certain cases you need to build out the tree from more than 1000 entries.

Best,
Thiago

walpolea:
I have wrapped Contentful with this type of function a few times, mostly for the management API. With the content API we've moved to the sync API and store all data in indexedDB so that the entire entry structure is present before rendering hierarchical structures. We've also considered having a node server proxy that syncs a contentful space with the sync API on a regular basis and serves us the entire cached JSON structure, which gzipped, can be pretty manageable in size to load. That in itself on the Contentful side could be a pretty nice feature. Not much else to add, but wanted it to be known that this use-case is not so close to the edge.

Best,
Andrew

Khaledgarbaya:
Hey guys,

We now have a method `parseEntiries` that accepts a raw json data and parse it into entries object with link resolution. Your part now is to get all the data and make sure always to concat both `data.items` and `data.include` for each page and then pass it to the client using `client.parseEntites(data)` and you will get a parsed data set with all the link resolution done for you.

Best,
Khaled

PR #90, opened 2016-09-09
feat(entities): add parseEntities method

Khaledgarbaya:
@stefanjudis could you take a look

### DIFF
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -196,27 +196,30 @@ export default function createContentfulApi ({http, shouldLinksResolve}) {
     return pagedSync(http, query, resolveLinks)
   }
   /**
-   * Parse a Given raw json data coming from the collection endpoint
-   * into collection of entry objects.
-   * Links will be resolved also
-   * @param {Object} data json data represting collection of entries
-   * @example
-   * let parsedData = client.parseEntries(data)
-   * console.log(parsedData)
-   * */
+  * Parse raw json data into collection of entry objects.
+  * Links will be resolved also
+  * @param {Object} data json data represting collection of entries
+  * @example
+  * <pre>
+  * let data = {};
+  * console.log( data.entries[0].fields.foo ); // undefined
+  * let parsedData = client.parseEntries(data);
+  * console.log( parsedData.entries[0].fields.foo ); // foo
+  * </pre>
+  * */
   function parseEntries (data) {
     return wrapEntryCollection(data, true, false)
   }
 
   return {
     getSpace: getSpace,
-    parseEntries: parseEntries,
     getContentType: getContentType,
     getContentTypes: getContentTypes,
     getEntry: getEntry,
     getEntries: getEntries,
     getAsset: getAsset,
     getAssets: getAssets,
+    parseEntries: parseEntries,
     sync: sync
   }
 }
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -202,9 +202,30 @@ export default function createContentfulApi ({http, shouldLinksResolve}) {
   * @example
   * <pre>
   * let data = {};
-  * console.log( data.entries[0].fields.foo ); // undefined
+  * const data = {items: [
+  *    {
+  *    sys: {type: 'Entry', locale: 'en-US'},
+  *    fields: {
+  *      animal: {sys: {type: 'Link', linkType: 'Animal', id: 'oink'}},
+  *      anotheranimal: {sys: {type: 'Link', linkType: 'Animal', id: 'middle-parrot'}}
+  *    }
+  *  }
+  * ],
+  * includes: {
+  *  Animal: [
+  *    {
+  *      sys: {type: 'Animal', id: 'oink', locale: 'en-US'},
+  *      fields: {
+  *        name: 'Pig',
+  *        friend: {sys: {type: 'Link', linkType: 'Animal', id: 'groundhog'}}
+  *      }
+  *    }
+  *   ]
+  *  }
+  * }
+  * console.log( data.items[0].fields.foo ); // undefined
   * let parsedData = client.parseEntries(data);
-  * console.log( parsedData.entries[0].fields.foo ); // foo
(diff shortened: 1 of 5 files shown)


lib/create-contentful-api.js
```diff
@@ -198,9 +195,22 @@ export default function createContentfulApi ({
     const resolveLinks = shouldLinksResolve(query)
     return pagedSync(http, query, resolveLinks)
   }
+  /**
+   * Parse a Given raw json data coming from the collection endpoint
```
stefanjudis: -> parse raw json data

Is maybe better.

lib/create-contentful-api.js
```diff
@@ -198,9 +195,22 @@ export default function createContentfulApi ({
...
+   * into collection of entry objects.
+   * Links will be resolved also
+   * @param {Object} data json data represting collection of entries
+   * @example
+   * let parsedData = client.parseEntries(data)
+   * console.log(parsedData)
```
stefanjudis: can we make it more clear what the power of this is?

it would be nice to see in the example that linking takes place. So for me to really understand it seeing an example json object and then calling a linked property would be awesome.

```
let data = {};

console.log( data.entries[0].fields.foo ); // undefined

let parsedData = client.parseEntries(data);

console.log( data.entries[0].fields.foo ); // foo
```

what do you think?

lib/create-contentful-api.js
```diff
@@ -198,9 +195,22 @@ export default function createContentfulApi ({
...
+   * */
+  function parseEntries (data) {
+    return wrapEntryCollection(data, true, false)
+  }
 
   return {
     getSpace: getSpace,
+    parseEntries: parseEntries,
```
stefanjudis: just preference - i wouldn't place `parse` in the middle of several `get`'s.

test/unit/create-contentful-api-test.js
```diff
@@ -352,3 +352,65 @@ test('CDA call sync fails', (t) => {
...
+      {
+        sys: {type: 'Animal', id: 'aussie-parrot', locale: 'en-US'},
+        fields: {name: 'Aussie Parrot'}
+      }
+    ]
+  }
+}
+  let parsedData = client.parseEntries(data)
+  t.ok(parsedData)
+  t.looseEquals(parsedData.items[0].fields.animal.sys, data.includes.Animal[0].sys, 'oink')
```
stefanjudis: When you're performing only one check here - you don't need so much example data, or?

so either less data or more checks would be good?

lib/create-contentful-api.js
Khaledgarbaya: Yeah sure, I will add a better example

stefanjudis:
also build's failing?

test/unit/create-contentful-api-test.js
```diff
@@ -352,3 +352,65 @@ test('CDA call sync fails', (t) => {
     teardown()
   })
 })
+
+test('Given json should be parsed correctly as a collection of entries', (t) => {
+  const client = contentful.createClient({accessToken: 'blablabla', space: 'bla'})
```
stefanjudis: I think we shouldn't do it like that here.

all the other test cases use something like

```
const {api} = setupWithData({
    promise: Promise.resolve({ data: contentTypeMock })
})
entitiesMock.contentType.wrapContentType.returns(contentTypeMock)

return api.getContentType('ctid')
```

for the matter of consistency and to have on less file to load ( 'require( '../../' )? ) it might make sense to use the same approach here.

Khaledgarbaya: That's completely different from what we are trying to achieve since we don't want to rewire the function we want the real logic to be executed

stefanjudis: then we should not rewire it, but use the same way of initializing the object?

I just think, that dropping in a new loaded file just for one test case, is not as clean as it should be.

The other test cases perform checks on methods that are right next to the new method we added e.g. ' getEntries'.

Maybe we can use this

```
const api = createContentfulApi({
  http: {
    get: getStub.returns(promise)
  },
  shouldLinksResolve: shouldLinksResolve
})
```

and use it without stubbing?

Reaching out to `../../` just feels like mixing the concept of this file up.

lib/create-contentful-api.js
```diff
@@ -198,6 +195,21 @@ export default function createContentfulApi ({
     const resolveLinks = shouldLinksResolve(query)
     return pagedSync(http, query, resolveLinks)
   }
+  /**
+  * Parse raw json data into collection of entry objects.
+  * Links will be resolved also
+  * @param {Object} data json data represting collection of entries
+  * @example
+  * <pre>
+  * let data = {};
```
stefanjudis: this example is misleading and would through an error, when I'd copy and paste it, as `entries` is not defined.

stefanjudis:
holy, coverage decreased by 17%?

Khaledgarbaya:
Something Wrong with travis ci, it executed the tests without the new test case, I fix that once i figure out the heck is going on with the ci

Khaledgarbaya:
Ok the build is failing now because of Sauce lab session which is not stable sometime

Khaledgarbaya:
All Green

test/unit/create-contentful-api-test.js
```diff
@@ -330,25 +336,58 @@ test('CDA call sync', (t) => {
...
+      teardown()
+    })
+})
+
+test('Given json should be parsed correctly as a collection of entries', (t) => {
+  const api = createContentfulApi({
+    http: {},
+    shouldLinksResolve: true
+  })
+  const data = {items: [
```
stefanjudis: Sweet! 👍

stefanjudis:
only this part missing https://github.com/contentful/contentful.js/pull/90#discussion_r78180549

and then we're ready to go i think. 👍

Khaledgarbaya:
done, but since it's Friday 😱 😱 😱 I will not merge till Monday

stefanjudis:
:+1

PR #108, opened 2017-01-13
umd build

Khaledgarbaya:
This PR is removing all legacy build setup and aim to have only one file for both targets node and browser to make integrating contentful.js with webpack for server side easier since we are droping support for old node versions. Also optimize lodash packaging.


.travis.yml
```diff
@@ -4,10 +4,9 @@ before_install:
   - npm set progress=false
 language: node_js
 node_js:
+  - "6.0"
```
axe312ger: we may already consider node 7?

Khaledgarbaya: the aim is to support the lts versions https://cloud.githubusercontent.com/assets/1156093/16716136/32f18fa6-46f4-11e6-8d56-3619807d38ad.png

axe312ger:
LGTM, the only concern I had i fixed on my own with the last commit. Works great with our discovery app :)

Merge them!

PR #111, opened 2017-01-23
fix: Bump babel-runtime dependency

ilkka:
This change bumps the babel-runtime dependency so that it in turn pulls in a newer minimatch. The version of minimatch used by this older babel-runtime is vulnerable to a regex DoS and causes vulnerability alerts from e.g. auditjs. Related babel PR: https://github.com/babel/babel/pull/3537

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "tonic-example.js"
   ],
   "dependencies": {
-    "babel-runtime": "~6.3.19",
+    "babel-runtime": "~6.22.0",
     "contentful-sdk-core": "~2.5.0",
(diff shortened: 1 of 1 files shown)


Khaledgarbaya:
@ilkka Thanks a lot, I'll just give the CI sometime because saucelab is timing out for now and merge it. I am working on umd build branch also here https://github.com/contentful/contentful.js/pull/108 it would be nice when you have time to give it a try.

Khaledgarbaya:
We are refactoring the whole SDK and this change is made already in another branch, I will close this PR and thank you for your cnotribution

Issue #114, opened 2017-02-05
Possibly incorrect documentation for createClient


sgronblo:
* **Node Version**:
* **Npm Version**:
* **Platform**:
* **Package Version**:

The docs for contentful 3.8.0 on https://contentful.github.io/contentful.js/contentful/3.8.0/contentful.html mention that the param to control the number of concurrent requests is called `concurrency`. However the function here https://github.com/contentful/contentful.js/blob/v3.8.0/lib/contentful.js#L41 passes `params.rateLimit` not `params.concurrency` to `wrapHttpClient` as `concurrency`. Params is also given to `httpCreateClient`, and I didn't dig all the way into contentful-sdk-core to check if those params are expected to contain `concurrency` instead of `rateLimit` in some other place.

Khaledgarbaya:
@sdepold Thanks for the report, I will take care of that asap, also PRs are welcome :)

xtianjohns:
I've opened a PR, @Khaledgarbaya, would like your feedback.

Khaledgarbaya:
@xtianjohns, LGTM. 👏

Khaledgarbaya:
I fixed the naming in the docs, sorry for the long delay https://contentful.github.io/contentful.js/contentful/3.8.1/contentful.html

PR #117, opened 2017-02-17
Add type definition file for typescript users

googol:
This enables typescript users to use the library in strongly typed mode without writing their own type definition files.

The types are based on the library documentation. However, some properties of the content types were not documented, such as the contentType-link on the Sys interface, and has been added.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,111 @@
+// Type definitions for contentful
+// Definitions by: Miika Hänninen <https://github.com/googol>
+
+export interface CreateClientParams {
+    space: string;
+    accessToken: string;
+    insecure?: boolean;
+    host?: string;
+    agent?: any;
+    headers?: any;
+    concurrency?: number;
+    delay?: number;
+    maxRetries?: number;
+    retryOnTooManyRequests?: boolean;
+    resolveLinks?: boolean;
+}
+
+export interface ContentfulClientApi {
+    getAsset(id: string, query?: any): Promise<Asset>;
+    getAssets(query?: any): Promise<AssetCollection>;
+    getContentType(id: string): Promise<ContentType>;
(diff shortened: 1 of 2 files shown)


Khaledgarbaya:
Hi  @googol , this is very cool thank a lot. I've been thinking about this for a long time but never had the time.

googol:
Feel free to ping me if there are any changes to the API and the typings need to be changed, I'm glad to help out with them

googol:
Btw @Khaledgarbaya, when can I expect a new release on npm with this included?

axe312ger:
@googol this week :)

theluk:
Hey,

somehow the d.ts file is still not included in the current package. 
in my node_modules your version.js file says 4.0.1 but no index.d.ts in folder can be found.

Khaledgarbaya:
@theluk sorry my bad, I should include the file in the package.json. will do it in few minutes

Khaledgarbaya:
@theluk the `index.d.ts` should be included now in the npm package `v4.0.2`

theluk:
thank you very much! 👍

sgronblo:
I forgot to say this when checking this PR initially but I think the `concurrency` property is the one that was incorrectly listed in the docs and the real name should have been `rateLimit` instead. As I reported earlier in this issue https://github.com/contentful/contentful.js/issues/114

Khaledgarbaya:
@sgronblo those params will be removed in the next release which should be soon. the SDK will handle rate limit automatically for you based on rate-limit headers provided by Contentful's endpoint.

sgronblo:
I understand that. But something which will be released "soon" does not help much if you are trying to use this library today. Aslo, is there some discussion going on somewhere about that change and the reasoning behind it?

Khaledgarbaya:
@sgronblo There will a detailed explanation about that. I will try to release a patched version with the correct documentation today.

PR #123, opened 2017-03-13
feat(rateLimit): Use the new ratelimit logic

Khaledgarbaya:
Some changes are required for to be able to use the new rate limiting introduced [here](https://github.com/contentful/contentful-sdk-core/pull/13)


googol:
Please remove the concurrency property from the type definition file if it is no longer needed with this change

googol:
:+1:

PR #127, opened 2017-03-23
fix(docs): `concurrency` -> `rateLimit` & `delay` -> `rateLimitPeriod`

Khaledgarbaya:


googol:
Please also change the type definition file or it will become out of sync

googol:
:+1:

PR #129, opened 2017-03-29
Update to the new docs template

Khaledgarbaya:
This PR fixes the documentation generation and updates to the new documentation template

screen shot 2017-05-30 at 14 13 04


Khaledgarbaya:
@axe312ger issue fixed

Khaledgarbaya:
@axe312ger can you elaborate on 
> I would love to have it with more contentful colors.

axe312ger:
Why not matching the style of our developer docs? https://www.contentful.com/developers/docs/references/content-management-api/

Change link color to the contentful blue, make the header black also, add contentful logo.

Did you change anything for mobile? It was pretty much broken last time I tried, but via Chrome Dev Tools Device Emulator it looks pretty fine now.

Khaledgarbaya:
> I guess generally most function could have proper parameter documentation, as is I did here: 
The know parameters are [there](https://github.com/contentful/contentful.js/blob/master/lib/contentful.js#L18)
For the linking to the functions I think we can use `@{link} tag` I will try it out

Khaledgarbaya:
I copy pasted the `colors` values from the web app I find it better but yeah we can match API docs to stay consistent

axe312ger:
<3

axe312ger:
Wonderful. Let's bring it online!

PR #130, opened 2017-03-29
Fix/umb build

Khaledgarbaya:
Thi PR fixes the export issue in the umd build for the browser


package.json
```diff
@@ -21,7 +22,7 @@
     "docs:watch": "watchy -w lib npm run docs:dev",
     "docs:publish": "npm run docs:build && ./node_modules/contentful-sdk-jsdoc/bin/publish-docs.sh contentful.js contentful",
     "lint": "eslint lib test",
-    "pretest": "npm run lint",
+    "pretest": "standard lib/*.js && standard lib/**/*.js && standard test/**/*.js",
```
axe312ger: nope nope, please do not remove eslint again

package.json
```diff
@@ -67,6 +67,7 @@
     "babel-loader": "^6.2.4",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-rewire": "^1.0.0",
+    "babel-plugin-transform-es2015-modules-umd": "^6.24.0",
```
axe312ger: not used also

version.js
```diff
@@ -1 +1 @@
-module.exports = ''
+module.exports = '4.1.0'
```
axe312ger: sure about this?

Khaledgarbaya: this is automated so it does not matter

PR #138, opened 2017-04-11
fixed "fields" member of ContentType interface

rossjp:
I want to be able to reference the `fields` array in my client-side code.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -64,7 +64,7 @@ export interface ContentType {
     name: string;
     description: string;
     displayField: string;
-    Array: string;
+    fields: Array;
     toPlainObject(): ContentType;
 }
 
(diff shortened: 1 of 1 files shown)


rossjp:
Unable to pass Sauce Labs tests due to encrypted environment variables being removed in Travis build.

googol:
Oh that's a pretty bad oversight in my original typings.
I'd like the array of fields to be strongly typed too. Would you mind me adding that to this PR?

index.d.ts
```diff
@@ -104,4 +104,14 @@ export interface ContentTypeLink {
     id: string;
 }
 
+export interface Field {
```
rossjp: How about something like this?

PR #142, opened 2017-04-20
Update TS typings for http agent and proxy config

googol:


googol:
It's a shame this didn't go in the latest release. Please consider that any public api change requires changes to the typings. Feel free to ping me if not sure

Khaledgarbaya:
@googol Sorry about that, I released this as a patch will be out in few minutes

googol:
Thanks!

Issue #145, opened 2017-05-08
`./es6-promise.map` polyfill warning with Webpack / TypeScript


Nevraeka:
The polyfill for Promise seems unreachable. Please advise on how to fix this warning. Thanks in advance.  cc @TheLarkInn

* **Node Version**: 6.8
* **Npm Version**: 3.10.8
* **Platform**: V-061G8WP 16.0.0 Darwin Kernel Version 16.0.0
* **Package Version**: contentful@4.3.0

#### Webpack output (with some altered paths & names)
```
WARNING in ./~/contentful/dist/contentful.js
(Emitted value instead of an instance of Error) Cannot find SourceMap 'es6-promise.map': Error: Can't resolve './es6-pro
mise.map' in '/node_modules/contentful/dist'
 @ ./app/utils/contentfulConnector.tsx 6:19-40
 @ ./app/containers/FeatureX/index.tsx
 @ ./app/containers/App/index.tsx
 @ ./app/app.tsx
 @ multi eventsource-polyfill webpack-hot-middleware/client?reload=true ./app/app.tsx
```

Khaledgarbaya:
Hi @Nevraeka,
Looks like you are using typescript.
can you share your `webpack.config` and `tsconfig`

Best,
Khaled

Nevraeka:
@Khaledgarbaya thank you 
#### TSCONFIG
```
{
    "compilerOptions": {
        "outDir": "./build/",
        "sourceMap": true,
        "noImplicitAny": false,
        "module": "commonjs",
        "moduleResolution": "node",
        "target": "es5",
        "jsx": "react",
        "experimentalDecorators": true,
        "removeComments": true,
        "allowJs": true,
        "lib": [
            "es2017",
            "dom"
        ],
        "typeRoots": [
            "node_modules/@types",
            "custom-typings",
            "typings/globals"
        ]
    },
    "awesomeTypescriptLoaderOptions": {
        "useWebpackText": true
    },
    "exclude": [
        "node_modules"
    ],
    "include": [
        "./app/**/*"
    ]
}
```
#### Webpack (base modified from react-boilerplate w/ custom scss modified)
```
/**
 * COMMON WEBPACK CONFIGURATION
 */

const path = require('path');
const webpack = require('webpack');
const { CheckerPlugin } = require('awesome-typescript-loader');

module.exports = options => ({
  entry: options.entry,
  output: Object.assign({
    path: path.resolve('build'),
    filename: 'app.js',
[...]

PR #146, opened 2017-05-10
fix(build): fix source-map and uglify

Khaledgarbaya:
This PR fixes two things:
1. we don't have to specify `-p` if we have `UglifyJsPlugin` inside the `webpack.config.js`
2. fix the `source-map` issue with es6-promises package mentioned [here](https://github.com/contentful/contentful.js/issues/145)

References : 
Webpack docs
>If --devtool source-map is omitted, the output still contains the sourcemap URI, but in an invalid position resulting in the browser ignoring it.


.gitignore
```diff
@@ -153,3 +153,4 @@ $RECYCLE.BIN/
 *.lnk
 
 # End of https://www.gitignore.io/api/node,windows,osx,linux,vim
+*.patch
```
axe312ger: Please add to beginning of the file as the custom entries

webpack.config.js
```diff
@@ -58,6 +59,7 @@ module.exports = [
     module: {
       loaders
     },
+    devtool: 'source-map',
```
axe312ger: For performance reasons, I would prefer `cheap-module-source-map` --> https://webpack.js.org/configuration/devtool/#devtool

Khaledgarbaya:
@axe312ger can you take a look again ?

aChudinov:
@Khaledgarbaya 
After this PR we have an issue with webpack UglifyJs
```
Error: app-bb3e1f83d765aff08857.js from UglifyJs
Unexpected token: operator (>) [./~/contentful/dist/contentful.js:1502,0][app-bb3e1f83d765aff08857.js:192762,17]
```

As I see, now we have ES6 code in compiled `dist` folder, like
```
const linkGetter = __WEBPACK_IMPORTED_MODULE_1_mem___default()(getLinksFromIncludes, {
  cacheKey: a => a.sys.id
});
```

PR #148, opened 2017-05-15
feature(createClient): Add Contentful User Agent Header

Khaledgarbaya:
### TODO

- [x] Add test
- [x]  Update index.d.ts
- [ ] Squash and merge


googol:
This requires a change to `index.d.ts` and the documentation as well

Khaledgarbaya:
@googol you are the guardian angel of the type definition. I added the 2 new params

index.d.ts
```diff
@@ -19,6 +19,8 @@ export interface CreateClientParams {
     httpsAgent?: any;
     proxy?: AxiosProxyConfig;
     headers?: any;
+    application: string;
```
googol: You'll want to add question marks after the names of application and integration, to make these optional.

lib/contentful.js
```diff
@@ -7,11 +7,10 @@
 
 import assign from 'lodash/assign'
 import axios from 'axios'
-import version from '../version'
 import {createHttpClient} from 'contentful-sdk-core'
 import createContentfulApi from './create-contentful-api'
 import createLinkResolver from './create-link-resolver'
-
+import {getUserAgentHeader} from './utils/user-agent'
```
axe312ger: It might be smart do add this function as a generic one to `contentful-sdk-core`?

axe312ger:
Still some errors with edge I think

Khaledgarbaya:
@axe312ger I fixed the Tests

PR #162, opened 2017-06-15
docs(readme): update to use const instead of var

petermakowski:
- updated to use const throughout the doc instead of var for consistency (const is used in newer sections of the doc and the rest of Contentful documentation)


Khaledgarbaya:
Thank you @petermakowski

Issue #164, opened 2017-06-20
Webpack build fails on contentful


alewaros:
* **Node Version**: 8.1.2
* **Npm Version**: 5.0.3
* **Platform**: macOS 10.12.5
* **Package Version**: 4.4.3

I'm importing contentful module on my webpack compiled project and when built, it produces these messages.

```
WARNING in ./node_modules/es6-promise/dist/es6-promise.js
Module not found: Error: Can't resolve 'vertx' in '/Users/aj/Developer/Selina/Selina-Booking-API/node_modules/es6-promise/dist'
 @ ./node_modules/es6-promise/dist/es6-promise.js 137:16-26
 @ ./node_modules/contentful-sdk-core/dist/es-modules/promised-wait.js
 @ ./node_modules/contentful-sdk-core/index.es-modules.js
 @ ./node_modules/contentful/dist/es6-modules/contentful.js
 @ ./src/index.js

ERROR in ./node_modules/contentful/dist/es6-modules/contentful.js
Module not found: Error: Can't resolve '../version.js' in '/Users/aj/Developer/Selina/Selina-Booking-API/node_modules/contentful/dist/es6-modules'
 @ ./node_modules/contentful/dist/es6-modules/contentful.js 13:0-36
 @ ./src/index.js
```

My webpack config file is this
```javascript
var path = require('path')

module.exports = {
  entry: './src/index.js',
  output: {
    filename: 'bundle.js',
    path: path.resolve(__dirname, 'dist')
  },
  module: {
    rules: [
      {
        test: /\.js$/,
        exclude: /node_modules/,
        use: ['babel-loader']
      }
    ]
  },
  target: 'node'
}
```

Khaledgarbaya:
Hi @alewaros,
I just pushed a PR to fix the issue https://github.com/contentful/contentful.js/pull/165 
can you check out that branch and test with it before I merge to master?

alewaros:
Sure thing

PR #165, opened 2017-06-21
fix(es6-modules): Add version to dist

Khaledgarbaya:
fixes https://github.com/contentful/contentful.js/issues/164


alewaros:
It works 👍

PR #172, opened 2017-07-17
house keeping

axe312ger:
* eslint to most current version
* upgrade to webpack 3 (cus it is zero hussle)
* introduce babeli
* clean up babel config, less babel polyfills in browser package!
* webpack config now finally uses const's


axe312ger:
Now comes with webpack-bundle-analyze and proper tap output:

screen shot 2017-07-19 at 14 59 42
screen shot 2017-07-19 at 15 01 46

PR #173, opened 2017-07-20
refactor(dependencies): replace lodash with smaller dependencies

axe312ger:
Based on https://github.com/contentful/contentful-sdk-core/pull/35

* Introduces object rest spread for a readable syntax of complex object assignments
* [mem](https://github.com/sindresorhus/mem) replaces `lodash.memoize`
* [clone](https://github.com/pvorb/clone) replaces `lodash.cloneDeep`
* [deep-get-set](https://github.com/acstll/deep-get-set) replaces `lodash.get`
* ES5 & ES6 replace the rest

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,22 +2,10 @@ const path = require('path')
 
 const webpack = require('webpack')
 const BabiliPlugin = require('babili-webpack-plugin')
-const LodashModuleReplacementPlugin = require('lodash-webpack-plugin')
 
 const PROD = process.env.NODE_ENV === 'production'
 
 const plugins = [
-  new LodashModuleReplacementPlugin({
-    'shorthands': true,
-    'cloning': true,
-    'currying': true,
-    'caching': true,
-    'collections': true,
-    'paths': true,
-    'guards': true,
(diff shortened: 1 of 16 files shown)


webpack.config.js
```diff
@@ -41,7 +29,11 @@ if (process.env.NODE_ENV === 'production') {
 const loaders = [
   {
     test: /\.js?$/,
-    exclude: /(node_modules|bower_components|dist)/,
+    include: [
+      path.resolve(__dirname, 'lib'),
+      path.resolve(__dirname, 'node_modules', 'mem'),
+      path.resolve(__dirname, 'node_modules', 'mimic-fn')
```
axe312ger: This is because mem & mimic-fn do not come as ES5

webpack.config.js
```diff
@@ -57,7 +49,8 @@ module.exports = [
       library: 'contentful'
     },
     module: {
-      loaders
+      loaders,
+      noParse: /\/clone\//
```
axe312ger: Avoids bundling of buffer polyfill with clone since we do not need buffer support an it is like 20kb big

PR #174, opened 2017-07-24
Next major version

axe312ger:
* [x] Upgrade dependencies
* [x] Reduce bundle size by removing/replacing dependencies
* [x] Improve dev tooling
* [x] Improve test tooling
* [x] Introduce a legacy version for IE11 and other older browsers (instead of forcing user to have a promise polyfill)
* [x] Introduce an es-modules version
* [x] Find an alternative to [platform.js](https://github.com/bestiejs/platform.js) since it is huge and breaks react-native
* [x] add missing configuration to readme. Proxy handling docs are missing/old.


axe312ger:
So missing atm compared to CMA branch:
* mem, get-deep-set and other should go back to lodash. Check https://github.com/contentful/contentful.js/pull/174/commits/c5ecb12ef7eee6ca8db162744797120d894c5c78 to find all of it
* webpack and babel config should be aligned to CMA
* make sure `test:only` is now `test:unit`
* upgrade to latest sdk-core
* adjust sauce labs configs, should be less failing with this settings :)
* get rid of version.js (not yet really tested, do the docs work? does it really include the correct version now?)

lib/mixins/link-getters.js
```diff
@@ -8,10 +12,8 @@ import mem from 'mem'
  * @param {Object} includes - Object with lists of Entry, Asset, DeletedEntry and DeletedAsset
  */
 export default function mixinLinkGetters (items, includes) {
-  const linkGetter = mem(getLinksFromIncludes, {
-    cacheKey: (a) => a.sys.id
-  })
-  items.forEach((item) => {
```
axe312ger: Keep the `forEach` to save `lodash.each`

lib/mixins/link-getters.js
```diff
@@ -29,8 +31,7 @@ export default function mixinLinkGetters (items, includes) {
     if (hasLocale) {
       setFieldGettersForFields(fields)
     } else {
-      const fieldKeys = Object.keys(fields)
-      fieldKeys.forEach((fieldKey) => setFieldGettersForFields(fields[fieldKey]))
```
axe312ger: Same as above

lib/mixins/link-getters.js
```diff
@@ -40,9 +41,7 @@ export default function mixinLinkGetters (items, includes) {
    * @param {Object} fields - Fields object
    */
   function setFieldGettersForFields (fields) {
-    const fieldKeys = Object.keys(fields)
-    fieldKeys.forEach((fieldKey) => {
-      const field = fields[fieldKey]
```
axe312ger: same

lib/mixins/link-getters.js
```diff
@@ -94,12 +93,15 @@ export default function mixinLinkGetters (items, includes) {
    * @return {Object} Field, or link if field not resolved
    */
   function getLinksFromIncludes (field) {
-    const linkedEntities = includes[field.sys.linkType] || []
-    const link = linkedEntities.find((linkedEntity) => linkedEntity.sys.id === field.sys.id)
```
axe312ger: `Array.find()` is fine, lets keep it :)

karma.conf.saucelabs.js
```diff
@@ -23,12 +23,13 @@ var browsers = {
 
 module.exports = function (config) {
   karmaBaseConf.plugins.push(require('karma-sauce-launcher'))
+  karmaBaseConf.plugins.push(require('karma-firefox-launcher'))
```
axe312ger: This is wrong, should be in the `karma.confg.base.js`

--> https://github.com/contentful/contentful-management.js/pull/127/files#diff-e3ca3f9c4ef50d2071076da33038a6fd
--> https://github.com/contentful/contentful-management.js/pull/127/files#diff-41f2621a7581f6a00915e53d89371399

axe312ger:
Readme is still missing a equivalent to https://github.com/contentful/contentful-management.js/blob/master/README.md#configuration

axe312ger:
@todo: check if we really need these babel dependencies since babel-preset-env might cover them:

```
    "babel-plugin-add-module-exports": "^0.2.1",
    "babel-plugin-transform-object-assign": "^6.22.0",
    "babel-plugin-transform-object-rest-spread": "^6.23.0",
    "babel-plugin-transform-runtime": "^6.7.5",
```

axe312ger:
@Khaledgarbaya Review needed for the following diff:
https://github.com/contentful/contentful.js/compare/a6cd41e4d0a441ad91cf1e23a14c9047ad1bb448...fd2e4e714891b67c26b20f8d4552124e321ed3cc

So far I can see, the rebase on master is the only missing step :)

Khaledgarbaya:
@axe312ger ci is failing

axe312ger:
Not anymore 😎

axe312ger:
Awesome. Will release the first RC tomorrow and then is time for dogfooding :)

package.json
```diff
@@ -76,6 +76,7 @@
     "blue-tape": "^1.0.0",
     "bundlesize": "^0.12.0",
     "contentful-sdk-jsdoc": "^2.2.0",
+    "core-js": "^2.5.1",
```
Khaledgarbaya: why bringing back `core-js`

axe312ger: Because it is a dependency of our code which was missing in one build I did. We should not rely on sub-dependencies. (It was babel in that case)

axe312ger: https://github.com/contentful/contentful.js/blob/next/webpack.config.js#L90-L96

PR #179, opened 2017-08-08
Fix sync link resolution

axe312ger:
the sync API sends deleted entries without fields. This broke the link resolution. This small bug fix fixes the issue :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,6 +253,9 @@ Your Space ID.
 #### host (default: `'cdn.contentful.com'`)
 Set the host used to build the request URI's.
 
+#### basePath (default: ``)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -253,6 +253,9 @@ Your Space ID.
 #### host (default: `'cdn.contentful.com'`)
 Set the host used to build the request URI's.
 
+#### basePath (default: ``)
```
Khaledgarbaya: why is this here ?

axe312ger: Because this also forces to upgrade to the latest sdk-core version which includes that feature

googol: It's missing from `index.d.ts`

axe312ger: @googol fixed 🙈  - we consider adding tests for this in the next version so it won't happen again :)

PR #182, opened 2017-08-21
fix(ie11): transform Object.assign also in sdk-core

### LINKED ISSUE #181: IE 11 Unable to load site when using package
* **Node Version**: 8.1.2
* **Npm Version**: 5.3.0
* **Platform**: Darwin Nick-Hess 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_6
* **Package Version**: 4.6.0

When using the Contentful package with my VueJS app I encounter an error on versions of the package newer than v3.5.0 with IE 11:

`Object doesn't support property or method 'assign'`

On v3.5.0 I do not encounter this error but on newer versions it occurs and none of my content on Contentful is loaded or displayed on the site.

axe312ger:
fixes #181

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -12,7 +12,8 @@
           },
           "modules": false
         }]
-      ]
+      ],
+      "plugins": ["transform-object-assign"]
     },
     "test": {
       "presets": [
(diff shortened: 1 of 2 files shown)


axe312ger:
Since babel is initialized differently in the saucelabs tests, this slipped through.

@Khaledgarbaya I think we should rewrite the saucelabs integration to use the real dist bundle and not the transpiled sourcecode directly?

axe312ger:
This means the CMA should also be broken for the IE11 :(

PR #193, opened 2017-10-20
docs(README): fix unpkg url and add non-node environment hint

### LINKED ISSUE #192: Example in README is incorrect
The example used to point to a specific version for browsers, like 4.1.1, does not work.
<script src="https://unpkg.com/contentful@4.1.1/dist/contentful.browser.min.js"></script>

Error:
Cannot find module "/dist/contentful.browser.min.js" in package contentful@4.1.1

axe312ger:
fixes #192

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,6 @@ If you'd like to use a standalone built file you can use the following script ta
 Using `contentful@latest` will always get you the latest version, but you can also specify a specific version number:
 
 ``` html
-<!-- Avoid using the following url for production. You can not rely on its availability. -->
 <script src="https://unpkg.com/contentful@5.0.1/dist/contentful.browser.min.js"></script>
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,17 +64,21 @@ yarn add contentful
...
+<!-- Avoid using the following url for production. You can not rely on its availability. -->
 <script src="https://unpkg.com/contentful@latest/dist/contentful.browser.min.js"></script>
 ```
 **It's not recommended to use the above URL for production.**
 
 Using `contentful@latest` will always get you the latest version, but you can also specify a specific version number:
 
 ``` html
-<script src="https://unpkg.com/contentful@4.1.1/dist/contentful.browser.min.js"></script>
+<!-- Avoid using the following url for production. You can not rely on its availability. -->
```
Khaledgarbaya: since we sepecify the exact version here we can remove this comment

README.md
```diff
@@ -87,6 +91,14 @@ This library also comes with a legacy version to support Internet Explorer 11 an
 
 To support legacy browsers in your application, use `contentful.legacy.min.js` instead of `contentful.browser.min.js`
 
+### Other environments:
```
Khaledgarbaya: Maybe let's move this section under troubleshooting section

PR #195, opened 2017-10-25
Remove git dependency for package axios.

drubin:
@axe312ger introduced this https://github.com/contentful/contentful.js/commit/6326f90dea771e7c1e68dfa44e2311a35d5ec4ad

This change broke a bunch of my docker builds that depend on contentful.

This changes it back to simple HTTP interface for fetches packages.

Our docker builder's don't have git and I would prefer not having to add git just for a temporary fix.

The syntax is slightly more verbose but allows dependencies to still be managed by vanilla npm instead of requiring git to be installed and configured.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "tonic-example.js"
   ],
   "dependencies": {
-    "axios": "contentful/axios#fix/https-via-http-proxy",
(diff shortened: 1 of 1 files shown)


drubin:
Thanks for merging this so quickly :)

PR #206, opened 2017-11-21
fix: fix type definition for sync payload

Khaledgarbaya:


index.d.ts
```diff
@@ -95,7 +95,7 @@ export interface Space {
 export interface SyncCollection {
     entries: Array<Entry<any>>;
     assets: Array<Asset>;
-    deletedEntries: Array<Entry<any>>;
+    deletedEntries: Array<any>;
     deletedAssets: Array<Asset>;
```
zcei: The same should be for `deletedAssets` - they won't resolve with the whole Asset, but only a meta-description

googol:
If deletedEntries and deletedAssets don't contain Entries/Assets, then what do they contain? Can that be modelled in a new interface?

The any type is likely too broad

zcei:
@googol yep, that's way better. It should be around this:

```ts
interface SpaceLink {
  id: string
  type: 'Link'
  linkType: 'Space'
}

enum SyncType {
  DeletedEntry = 'DeletedEntry',
  DeletedAsset = 'DeletedAsset'
}

interface SyncDeletedEntity {
  sys: {
    space: SpaceLink
    id: string
    type: SyncType
    createdAt: string
    updatedAt: string
    deletedAt: string
    revision: number
  }
}
```

I just figured, that the `Sys` interface always has a property `contentType`, which doesn't match all cases. Maybe this should be turned into some kind of generic, or have multiple different `Sys` interfaces? Same for Links, we currently have `ContentTypeLink` and this would now introduce a `SpaceLink`

Khaledgarbaya:
Alright I will change that asap

googol:
Different Sys interfaces would be ideal. Let me know if I can help!

index.d.ts
```diff
@@ -95,7 +95,7 @@ export interface Space {
...
     nextSyncToken: string;
     toPlainObject(): SyncCollection;
```
grub203: 7248068f53cf331d6dca4cd6f4d057926dd6483f

makinwab:
Closing this PR as it's been open for a couple of years now. If there's a very much needed change we need to incorporate, please feel free to reopen.

PR #216, opened 2018-01-03
feat(links): unresolved links are ignored

Khaledgarbaya:
This PR changes the link resolution, any non-resolved link should be removed from the payload object.
Still not sure if this should be a breaking change or not

### DIFF
--- a/lib/entities/entry.js
+++ b/lib/entities/entry.js
@@ -100,7 +100,7 @@ export function wrapEntryCollection (data, resolveLinks, resolveForAllLocales) {
   const wrappedData = mixinStringifySafe(toPlainObject(cloneDeep(data)))
   if (resolveLinks) {
     const includes = prepareIncludes(wrappedData.includes, wrappedData.items)
-    mixinLinkGetters(wrappedData.items, includes, resolveForAllLocales)
+    mixinLinkGetters(wrappedData.items, includes)
   }
   return freezeSys(wrappedData)
 }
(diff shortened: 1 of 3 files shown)


lib/entities/entry.js
```diff
@@ -100,7 +100,7 @@ export function wrapEntryCollection (data, resolveLinks, resolveForAllLocales) {
   const wrappedData = mixinStringifySafe(toPlainObject(cloneDeep(data)))
   if (resolveLinks) {
     const includes = prepareIncludes(wrappedData.includes, wrappedData.items)
-    mixinLinkGetters(wrappedData.items, includes, resolveForAllLocales)
+    mixinLinkGetters(wrappedData.items, includes)
```
axe312ger: Shouldn't we pass `wrappedData.errors` as last argument?

lib/mixins/link-getters.js
```diff
@@ -40,7 +40,10 @@ export default function mixinLinkGetters (items, includes) {
     const fieldKeys = Object.keys(fields)
     fieldKeys.forEach((fieldKey) => {
       const field = fields[fieldKey]
-      if (Array.isArray(field)) {
+      const unresolved = errors.find((error) => error.details.id === field.sys.id)
```
axe312ger: Is it possible to have different errors besides resolving errors?

If yes, how do we wanna handle them? Just remove all fields with errors? (Sounds good to me, just want to clarify)

PR #217, opened 2018-01-04
fix(links): fix circular reference link resolution

Khaledgarbaya:
This PR removes the old link resolutions logic instead it uses https://github.com/contentful/contentful-resolve-response.
Now the SDK can resolve links even circular dependencies, also you can set `removeUnresolved` if a user wants to remove all unresolved links, this is useful if you have drafts items.


Khaledgarbaya:
Object destructuring is not supported for node 4.7

lib/create-contentful-api.js
```diff
@@ -56,10 +56,10 @@ import pagedSync from './paged-sync'
  * @param {Object} params - API initialization params
  * @prop {Object} http - HTTP client instance
  * @prop {Object} entities - Object with wrapper methods for each kind of entity
- * @prop {Function} shouldLinksResolve - Link resolver preconfigured with global setting
+ * @prop {Function} getGlobaloptions - Link resolver preconfigured with global setting
```
axe312ger: typo: getGlobal__o__ptions

lib/paged-sync.js
```diff
@@ -56,7 +56,8 @@ export default function pagedSync (http, query, resolveLinks) {
     .then((response) => {
       // clones response.items used in includes because we don't want these to be mutated
       if (resolveLinks) {
-        mixinLinkGetters(response.items, mapIncludeItems(cloneDeep(response.items)))
+        response.includes = mapIncludeItems(cloneDeep(response.items))
+        response.items = resolveResponse(response, {resolveLinks, removeUnresolved})
```
axe312ger: No need for this cloneDeep since resolve-response is already doing it:
https://github.com/contentful/contentful-resolve-response/blob/master/index.js#L92

package.json
```diff
@@ -27,11 +27,11 @@
...
-    "test:cover": "BABEL_ENV=test babel-node ./node_modules/istanbul/lib/cli.js cover ./test/runner | tap-spec",
-    "test:unit": "BABEL_ENV=test babel-node ./test/runner | tap-spec",
-    "test:debug": "BABEL_ENV=test babel-node debug ./test/runner",
-    "test:integration": "BABEL_ENV=test babel-node ./test/integration/tests.js | tap-spec",
-    "test:integration-debug": "BABEL_ENV=test babel-node debug ./test/integration/tests.js",
+    "test:cover": "BABEL_ENV=test node ./node_modules/istanbul/lib/cli.js cover ./test/runner | tap-spec",
+    "test:unit": "BABEL_ENV=test node ./test/runner | tap-spec",
+    "test:debug": "BABEL_ENV=test node debug ./test/runner",
+    "test:integration": "BABEL_ENV=test node ./test/integration/tests.js | tap-spec",
+    "test:integration-debug": "BABEL_ENV=test node debug ./test/integration/tests.js",
```
axe312ger: 🚀  👍

test/runner
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env node
+require('babel-core/register')({
+  // This will override `node_modules` ignoring - you can alternatively pass
+  // a regex
```
axe312ger: ` - you can alternatively pass a regex`

copy 🍝 ?

test/unit/create-contentful-api-test.js
```diff
@@ -379,7 +379,7 @@ test('CDA call sync fails', (t) => {
 test('Given json should be parsed correctly as a collection of entries', (t) => {
   const api = createContentfulApi({
     http: {},
-    shouldLinksResolve: true
+    getGlobalOptions: sinon.stub().returns({resolveLinks: true})
```
axe312ger: A check is missing here if `getGlobalOptions` was actually called

OR just remove the stub and use a function

axe312ger:
One more comment 👆 

Plus still:

> I can not locate the test which is testing the settings overwrite by query parameters.
> Can you please point me to the test and the docs where this is mentioned? If one or both are missing, we should add it.

Khaledgarbaya:
@axe312ger the tests are [here](https://github.com/contentful/contentful.js/blob/master/test/unit/contentful-test.js#L74-L102)

axe312ger:
@Khaledgarbaya nope, these set the resolution as global config.

I am talking about overwriting that config per query. Like `client.getEntries({resolveLinks: false})` or sth similar

Khaledgarbaya:
@axe312ger Added test

PR #226, opened 2018-02-12
fix(link-resolution): only resolve links in fields

axe312ger:
Current version of the SDK resolves links within `sys` if `removeUnresolved: true`

This fixes this


axe312ger:
closing in favour of #227

PR #241, opened 2018-04-25
Add missing CreateClientParams type definitions

### LINKED ISSUE #239: TypeScript definition for CreateClientParams is missing supported field/fields
* **Package Version**: 5.1.3

The definition are missing the parameter `timeout` which required me to extend the typings:

```typescript
export interface ContentfulCreateClientParams extends CreateClientParams {
    timeout?: number;
}
```

Possibly some other fields are missing too?

ristomatti:
This adds missing `CreateClientParams` TypeScript type definitions and  fixes #239.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,9 +10,12 @@ export interface AxiosProxyConfig {
     };
 }
 
+export type ClientLogLevel = 'error' | 'warning' | 'info';
+
 export interface CreateClientParams {
     space: string;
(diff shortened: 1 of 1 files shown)


axe312ger:
Looks good but I won't approve since I am not using typescript at the moment.

Can @googol maybe have a look? :)

Khaledgarbaya:
Thanks a lot @ristomatti

PR #244, opened 2018-05-07
docs: Add instruction about Angular 6

Khaledgarbaya:

PR #248, opened 2018-05-18
Revert "docs: Add instruction about Angular 6"

zcei:
Reverts contentful/contentful.js#244

After #247 landed, these instructions are no longer needed.


zcei:
Wat.. why does it also revert my fix?! I just used the `Revert` button on #244 🙁

zcei:
Seems to be a bug, the commit is empty, only the docs would be reverted

PR #253, opened 2018-07-09
Test/fix configuration

phoebeschmidt:
## Summary

* Fixed karma config so tests run with new webpack version. 
* Bumped `bundlesize` version and increased bundlesize limits in `package.json`


axe312ger:
👌! Lets 🚢 it.

Issue #208, opened 2017-11-27
Typescript Asset type error


rewop:
The typescript Asset is wrong. When I call the `client.sync` method the following object for the asset is returned:

```js
{ 
  sys: { 
     space: { sys: [Object] },
     type: 'Asset',
     id: 'xxxx',
     revision: 0,
     createdAt: '2017-11-01T15:36:18.377Z',
     updatedAt: '2017-11-01T15:36:21.510Z' 
  },
  fields: { 
     title: { en: 'xxxxxxx' },
     file: { en: { file: { url: 'xxxxx' } } } 
  } 
}
```

The `fields` props are localized (we need to acces en), while the definition of the ts type is not:
https://github.com/contentful/contentful.js/blob/master/index.d.ts#L46

Khaledgarbaya:
Hey @rewop this will be fixed in https://github.com/contentful/contentful.js/pull/206

marcolink:
Hey @rewop,
Good catch, the shape of the `sync` result differs indeed from the one for `getEntries`.
Do you have a chance to open a pull request for it?

Issue #143, opened 2017-04-21
Create tests for typescript type definition


Khaledgarbaya:
It would be nice to have some tests for typescript type definition file to make sure it is always in sync

stefanjudis:
I'm playing around with typescript right. Do you have a clue already how you would do this? :)

axe312ger:
@stefanjudis Khaled is in holidays and I used TS a few times but never created my own typings. So, nope.

Maybe @googol can assist here?

mgeorgehansen:
I just submitted a PR (#258) with a potential testing solution using [ts-runtime](https://github.com/fabiandev/ts-runtime). If there's interest in developing this into a production-ready state please let me know and I'd be happy to collaborate on it!

marcolink:
Related PR has been closed. will close this issue as well.

PR #258, opened 2018-07-22
[WIP] TypeScript declaration tests

mgeorgehansen:
## Summary

Adds tests that check the TypeScript declarations against the runtime behavior of this library.

## Description

The TypeScript declarations for contentful.js currently have no automated testing process to verify them, which makes keeping them in sync with the js codebase a manual testing process.

This PR adds automated tests of the TypeScript declarations against the runtime behavior of the js codebase using [fabiandev/ts-runtime](https://github.com/fabiandev/ts-runtime), an experimental but seemingly functional library based on [codemix/flow-runtime](https://github.com/codemix/flow-runtime). ts-runtime takes TypeScript files with explicit variable type annotations and compiles them into js files with runtime assertions about the shape of annotated variables.

Currently the source TypeScript files under test/typings are compiled into typings tests and placed into build/typings-tests. The scripts/test-typings.js script runs those tests and adapts their results to blue-tape test cases.

This also adds a few helper commands to package.json: "clean:typings-tests", "build:typings-tests", and "test:typings".

Tests currently use the generated NodeJS version of contentful.js, so run "npm run build:standalone" first before running the typing tests.

The existing tests have already identified a few typing mismatches:

1. ContentfulClientAPI.getEntries() strips the Entry<T>.toPlainObject() function from the entities and thus does not conform to the TypeScript definition
2. ContentfulClientAPI.getAssets() also strips the Asset.toPlainObject() function
3. Asset.fields.description is nullable but not declared as such
4. Asset.sys.contentType is declared but does not exist (seems to relate to discussion of Sys interface at https://github.com/contentful/contentful.js/pull/206#issuecomment-346590797)

## Motivation and Context

This PR primarily addresses #143, but is also motivated by other open issues with TypeScript declaration mismatches (e.g. #208). Having automated tests of the TypeScript declarations can lead to quicker identification of type mismatches and help prevent regressions.

## Todos

- [ ] Discuss/research if this is the best approach
- [ ] Test on Node v6, v8 (only been tested on Node v10)
- [ ] Add a few more tests (client.getAsset, client.getcontentType(s), client.getSpace, client.sync)


axe312ger:
Wonderful, thank you. I will have a closer look soon.

I wonder what @googol thinks, since he contributed a lot of our current typescript definitions.

makinwab:
Closing this PR as it's been open for a couple of years now. If there's a very much needed change we need to incorporate, please feel free to reopen.

PR #261, opened 2018-08-06
Housekeeping

axe312ger:


Khaledgarbaya:
If we're gonna release a beta version the base of this PR should be `next` oder?

axe312ger:
Why? The git branch name is irrelevant, just the npm tag needs to be next

Issue #260, opened 2018-07-26
Remove browser field from package.json


ravikiran438:
The use of the browser field in package.json is superfluous as it just points to a bundled file of all the modules and results in the inclusion of contentful.browser.js in the final webpack bundle thereby defeating the purpose of using es6 modules for tree shaking. The purpose of the browser field is to provide code that is specific to a browser (for example it uses polyfills or shims), but that is not relevant for this lib at present. For spec of browser field see: https://github.com/defunctzombie/package-browser-field-spec

## Steps to Reproduce
1) Launch bundle analyzer with release tag 6.1.0 and then verify with tag 6.1.1

Here is the relevant PR https://github.com/contentful/contentful.js/pull/247
Also same with contentful-management
https://github.com/contentful/contentful-management.js/pull/171

axe312ger:
Which might break Angular 6 again :/

We need to think this through first and then come up with a solution asap. Thanks for the report @ravikiran438

ravikiran438:
@axe312ger ,

As mentioned in https://github.com/angular/angular-cli/issues/10681#issuecomment-396848967

adding 

  "browser": {
    "os": false
  },

in contentful-sdk-core or using 

https://github.com/CoderPuppy/os-browserify

both seem to work fine with angular 6 builds

I can send out a PR if you would like to add any of those workarounds.

Thanks!

Khaledgarbaya:
Hey @ravikiran438,
Please feel free to send a PR
Best,
Khaled

ravikiran438:
@Khaledgarbaya Please check the following PR https://github.com/contentful/contentful-sdk-core/pull/62

PR #262, opened 2018-08-06
Fix: Remove browser field as contentful-sdk-core has been fixed now

ravikiran438:
## Summary

We can safely remove the browser field now thus fixing the webpack tree shaking. See relevant issue:
https://github.com/contentful/contentful.js/issues/260


axe312ger:
Thank you. The new version should be out in no time :)

PR #263, opened 2018-08-07
Feat/get entry to collection endpoint

phoebeschmidt:
## Summary

Use collection endpoint for getEntry to take advantage of link resolution.

### DIFF
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -160,10 +160,13 @@ export default function createContentfulApi ({http, getGlobalOptions}) {
    * .catch(console.error)
    */
   function getEntry (id, query = {}) {
-    switchToEnvironment(http)
-    normalizeSelect(query)
-    return http.get('entries/' + id, createRequestConfig({query: query}))
-      .then((response) => wrapEntry(response.data), errorHandler)
+    return this.getEntries({'sys.id': id, ...query})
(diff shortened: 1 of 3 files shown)


lib/create-contentful-api.js
```diff
@@ -160,10 +160,13 @@ export default function createContentfulApi ({http, getGlobalOptions}) {
...
-    switchToEnvironment(http)
-    normalizeSelect(query)
-    return http.get('entries/' + id, createRequestConfig({query: query}))
-      .then((response) => wrapEntry(response.data), errorHandler)
+    return this.getEntries({'sys.id': id, ...query})
+      .then((response) => {
+        if (response.items.length > 0) {
+          return wrapEntry(response.items[0])
+        }
+        throw new Error(`Entry with id ${id} cannot be found.`)
```
Khaledgarbaya: Let's throw the same error as before for consistency:

we can throw an object like this maybe
```js
{
  "sys": {
    "type": "Error",
    "id": "NotFound"
  },
  "message": "The resource could not be found.",
  "details": {
    "type": "Entry",
    "id": id,
    "environment": getGlobalOptions().environment,
    "space": getGlobalOptions().space
  }
}
```

phoebeschmidt: 👍 yes good call

PR #266, opened 2018-08-14
Add getLocales and Locales interface to index.d.ts

francescostella:
## Summary

Adding the missing method `getLocales()` to the Contentful client interface

## Description

Added:
- `getLocales()` method (documented and implemented https://contentful.github.io/contentful.js/contentful/7.0.0/ContentfulClientAPI.html#.getLocales) but not declared in the `index.d.ts`
- `Locales` interface

## Motivation and Context
If the Contentful library will be used with a Typescript compiler it will throw errors because the declaration of the type is missing from the `index.d.ts` file.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,7 +40,7 @@ export interface ContentfulClientApi {
     getEntries<T>(query?: any): Promise<EntryCollection<T>>;
     getEntry<T>(id: string, query?: any): Promise<Entry<T>>;
     getSpace(): Promise<Space>;
-    getLocales(): Promise<Locales>;
+    getLocales(): Promise<LocaleCollection>;
     sync(query: any): Promise<SyncCollection>;
 }
 
@@ -99,15 +99,6 @@ export interface Space {
     toPlainObject(): Space;
 }
 
-export interface Locales {
-    sys: Sys;
-    total: number;
-    skip: number;
-    limit: number;
-    items: ContentfulCollection<Locale>;
-    toPlainObject(): Space;
-}
-
 export interface Locale {
   code: string
   name: string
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -98,6 +99,15 @@ export interface Space {
     toPlainObject(): Space;
 }
 
+export interface Locales {
+    sys: Sys;
+    total: number;
+    skip: number;
+    limit: number;
+    items: Array<string>;
```
TimBeyer: Shouldn't this return a `Locale` type - which currently does not exist - or `any`?

I think the Locale type would look something like

```
interface Locale {
  code: string
  name: string
  default: boolean
  fallbackCode: string | null
  sys: {
    id: string
    type: 'Locale'
    version: number
  }
}
```

Khaledgarbaya: This should be `ContentfulCollection<Locale>` Locales should be removed and as Tim mentioned we should add Locale interface

```js
interface Locale {
  code: string
  name: string
  default: boolean
  fallbackCode: string | null
  sys: {
    id: string
    type: 'Locale'
    version: number
  }
}
```

francescostella: Thanks @Khaledgarbaya and @TimBeyer for the feedback and directions.
I addressed your feedback and pushed the changes to this PR.

index.d.ts
```diff
@@ -98,6 +99,27 @@ export interface Space {
     toPlainObject(): Space;
 }
 
+export interface Locales {
```
Khaledgarbaya: Sorry for the confusion, we don't need to have ´Locales´ Type since we have ContentfulCollection.
The type definition for locales should be like the following.

```js
getLocales(): Promise<LocaleCollection>;

// ....

export interface Locale {
  code: string
  name: string
  default: boolean
  fallbackCode: string | null
  sys: {
    id: string
    type: 'Locale'
    version: number
  }
}

export type LocaleCollection = ContentfulCollection<Locale>;
```

francescostella:
@Khaledgarbaya 👍 got it! please take a look, just updated it accordingly

PR #267, opened 2018-08-16
docs(ADVANCED.md): Update link resolution description

phoebeschmidt:
## Summary

Update the description of link resolution in ADVANCED.md to include differences between v7.0.0 (resolves links for `getEntry()`) and older versions (only collections endpoint resolves links).

### DIFF
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -41,7 +41,7 @@ const client = contentful.createClient({
 
 Link resolution occurs on the `getEntry` and `getEntries` endpoints as of version 7.0.0. For previous versions, only the collections endpoint resolved links. See [note](#note:-link-resolution-for-versions-older-than-7.0.0) below for more details.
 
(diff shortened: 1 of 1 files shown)


ADVANCED.md
```diff
@@ -38,7 +39,15 @@ const client = contentful.createClient({
 });
 ```
 
-Please note that the link resolution is only possible when requesting records from the collection endpoint using `client.getEntries()` or by performing an initial sync `client.sync({initial: true})`. In case you want to request one entry and benefit from the link resolution you can use the collection end point with the following query parameter `'sys.id': '<your-entry-id>'`.
+Link resolution occurs on the `getEntry` and `getEntries` endpoints as of version 7.0.0. For previous versions, only the collections endpoint resolved links. See [note](#note:-link-resolution-for-versions-older-than-7.0.0) below for more details.
+
+The link resolution is applied to one level deep by default. If you need it to be applied deeper, you may specify the `include` parameter when fetching your entries as follows `client.getEntries({include: <value>})`. The `include` parameter can be set to a number up to 10..
```
axe312ger: one dot to many at the end

phoebeschmidt: fixed!

PR #270, opened 2018-08-24
Make arguments Optional for Stringify safe method

celestius0:
TS linter complains when you try to call stringifySafe without args as suggested here:

https://www.contentful.com/developers/docs/javascript/tutorials/using-the-sync-api-with-js/

## Todos

-   [x] Implemented feature


Khaledgarbaya:
Thank you @celestius0

PR #274, opened 2018-09-03
Babel v7

axe312ger:
Seems to work fine, simplifies a lot of our config.

Just.. require-all, tape, and related tools seem to hate babel upgrades. We need jest :tm:


Algram:
Obsolete by https://github.com/contentful/contentful.js/pull/342

PR #281, opened 2018-09-26
fix(Sync): add toPlainObject method to Entries and Assets

### LINKED ISSUE #280: Sync API Entries Not of Type Entry
## Expected Behavior

```javascript
client.sync({ initial: true })
  .then(response => {
    response.entries.forEach(entry => {
      const plain = entry.toPlainObject() // <- This should be an Entry?
```
I am using the 7.1.0-beta0 SDK according to your docs:

- `response` ([link](https://contentful.github.io/contentful.js/contentful/7.1.0-beta0/ContentfulClientAPI.html#.sync)) should a `SyncCollection`
- `response.entries` ([link](https://contentful.github.io/contentful.js/contentful/7.1.0-beta0/Sync.html#.SyncCollection) should be an `Array.<Entities.Entry>`
- `entry` ([link](https://contentful.github.io/contentful.js/contentful/7.1.0-beta0/Entities.html#.Entry)) should now be an `Entry`

## Actual Behavior

The `entry` variable doesn't have a `toPlainObject` on it, and I'm not able to use this object to store it anywhere since it contains circular references.

## Possible Solution

I've already tried to import `toPlainObject` from contentful-sdk-core and pass the data in though this adds the `toPlainObject` function on the object passed into it, calling it still maintains the circular references.

## Steps to Reproduce

I think simply using the client's sync API call, grabbing entries from the response and the first entry more specifically should be enough to reproduce that the object doesn't have a `toPlainObject` on it.

## Environment

-   **Language Version**:  v8.9.1
-   **Package Manager Version**: yarn 1.9.4
-   **Operating System**:  Darwin machine.local 16.7.0 Darwin Kernel Version 16.7.0
-   **Package Version**:  7.1.0-beta0 and 7.0.3
-   **Which API are you using?**:  Delivery

phoebeschmidt:
fixes #280

### DIFF
--- a/lib/paged-sync.js
+++ b/lib/paged-sync.js
@@ -93,11 +93,20 @@ export default function pagedSync (http, query, options = {}) {
  * @return {Object} Entities mapped to an object for each entity type
  */
 function mapResponseItems (items) {
+  const reducer = (type) => {
+    return (accumulated, item) => {
+      if (item.sys.type === type) {
+        accumulated.push(toPlainObject(item))
+      }
+      return accumulated
+    }
+  }
+
   return {
-    entries: items.filter((item) => item.sys.type === 'Entry'),
-    assets: items.filter((item) => item.sys.type === 'Asset'),
-    deletedEntries: items.filter((item) => item.sys.type === 'DeletedEntry'),
-    deletedAssets: items.filter((item) => item.sys.type === 'DeletedAsset')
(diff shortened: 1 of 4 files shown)


phoebeschmidt:
FYI: upgrading webpack CLI fixed strange error causing CI to fail: https://github.com/webpack/webpack/issues/8082

PR #301, opened 2019-01-17
chore(types): Asset generic type for file details property

vmasek:
## Summary

Improved typing for `Asset` `fields.file.details` property

## Description

Used type generics with defauld of `any` to represent the property

## Motivation and Context

It will provide more convinient type checking for Asset's details property

-   [x] Implemented feature

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,14 +44,14 @@ export interface ContentfulClientApi {
     sync(query: any): Promise<SyncCollection>;
 }
 
-export interface Asset {
+export interface Asset<T = any> {
     sys: Sys;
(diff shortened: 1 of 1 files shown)


jarkkosyrjala:
I'm not sure what is your use case, but i found that the details property is pretty generic with all assets.

```
            details: {
                size: number;
                image?: {
                    width: number; 
                    height: number;
                };
            };
```
I created a PR that adds that to typings at https://github.com/contentful/contentful.js/pull/305

vmasek:
If the image is the only asset with "special" property, then yes your PR should be enough

PR #314, opened 2019-03-11
feat(dependencies): bump to beta version of axios instead of @contentful/axios

phoebeschmidt:
We have been managing many different versions of our sdks that point to this Axios beta instead of our (almost) out of date @contentful/axios for a while now. 

There are many reasons to depend on the beta, including (but not limited to) the way axios handles proxy configs in the newest beta. Also several Contentful tools already rely on the axios beta in their master releases.


package.json
```diff
@@ -58,7 +58,7 @@
     "tonic-example.js"
   ],
   "dependencies": {
-    "@contentful/axios": "^0.18.0",
+    "axios": "^0.19.0-beta.1",
     "contentful-resolve-response": "^1.1.4",
     "contentful-sdk-core": "^6.2.1",
```
Khaledgarbaya: can you update this to `6.3` ?

PR #329, opened 2019-06-23
fix: fix EntryFields.RichText types

nju33:
## Summary

🤔Looks like the type is wrong.

[Image from Gyazo](https://gyazo.com/ccdb8f7d68f5f9b9ae25871bc420fe9b)

Its `content` is really array but type is object.

## Description

I changed it into array.


sergei-matheson:
👍 Would love to have this. Currently causing errors on our project.

PR #368, opened 2020-01-27
fix(types) add revision and space to Sys interface

### LINKED ISSUE #351: Sys interface in typings is missing revision property
The typings are missing the property `revision`, which is a `number` when an entry is published, and `undefined` when it is not.

https://github.com/contentful/contentful.js/blob/559733a47b2dcf8918f46c15f7d10ae366f09e49/index.d.ts#L135-L144

This should be changed so that the typings contain the `revision` property.

pkhpkh:
## Summary

Changes the `Sys` definition in types to include `revision` and `space` fields.

## Description

Changes the `Sys` definition in types to include `revision` and `space` fields. They're included optionally -- `revision` is actually optional, since it's `undefined` for unpublished content. I left `space` as optional in order to avoid a breaking change. The type for `space` is defined similarly to the existing definition for `contentType`.

`contentType` should also probably be optional (since it's not set on the `sys` field for content types), but I wasn't entirely sure on that.

## Motivation and Context

Fixes #351

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -139,9 +139,19 @@ export interface Sys {
     createdAt: string;
     updatedAt: string;
     locale: string;
-    contentType: {
+    revision?: number;
+    space: {
+        sys: SpaceLink;
+    };
+    contentType?: {
         sys: ContentTypeLink;
     };
+}
+
+export interface SpaceLink {
+    type: 'Link';
+    linkType: 'Space';
+    id: string;
 }
 
 export interface ContentTypeLink {


chchrist:
Hello, I would like to inform you that this release has broken our ci/cd pipelines. The part that made the contentType optional. We are using Contentful JS SDK from a node.js service. We fixed that but not sure we can protect from breaking releases in the future

btakashi:
+1 @chchrist , the change of contentType to optional also broke my build

tomimaen:
This is breaking our build as well, in our project we are actually tripping on the part where the new `space` field **should** be optional (as it is written in the description that this was left to be optional to avoid a breaking change) but in the typings it actually is required.

PR #374, opened 2020-02-20
Delete limit from sync token

jordan314:
The Sync API has brand new 'limit' parameter. It can only be set when initial: true, and should not be sent with syncTokens. This deletes the limit from the sync query if a sync_token is set, otherwise you get the error: When passing a sync token you can not pass any other parameters. If you want to sync with a specific limit, start a new sync with "initial".

## Summary

The error `When passing a sync token you can not pass any other parameters. If you want to sync with a specific limit, start a new sync with "initial".`

## Description

The Sync API has brand new 'limit' parameter. It can only be set when initial: true, and should not be sent with syncTokens. This deletes the limit from the sync query if a sync_token is set, otherwise you get the error: When passing a sync token you can not pass any other parameters. If you want to sync with a specific limit, start a new sync with "initial".

## Motivation and Context

Fixes the above error

## Todos

-   [x] Implemented feature

### DIFF
--- a/lib/paged-sync.js
+++ b/lib/paged-sync.js
@@ -137,6 +137,7 @@ function getSyncPage (http, items, query, { paginate }) {
   }
 
   if (query.sync_token) {
+    delete query.limit
     delete query.initial
     delete query.type
     delete query.content_type


makinwab:
Hi @jordan314, PR #373 I believe also addresses the same issue?

jordan314:
Yeah

PR #389, opened 2020-04-02
chore(config): Add dependabot config and update semantic-release settings

marcolink:


package.json
```diff
@@ -145,7 +145,25 @@
     },
     {
       "path": "./dist/contentful.node.min.js",
-      "maxSize": "21Kb"
+      "maxSize": "22Kb"
```
TimBeyer, line 148: 🙃

Issue #371, opened 2020-02-05
contentful.legacy.min.js not working on IE 11


kingeric027:
## Expected Behavior
I am using the legacy contentful library described here https://contentful.github.io/contentful.js/contentful/6.1.0/

my script tag in my index html is 
`<script src='https://cdn.jsdelivr.net/npm/contentful@latest/dist/contentful.legacy.min.js'></script>`

I would expect this to work on IE 11:
"To support legacy browsers in your application, use contentful.legacy.min.js instead of contentful.browser.min.js"

## Actual Behavior

The app does not load on IE 11. I get this console error: 
SCRIPTING1006: Expected ')'
contentful.legacy.min.js (1,43092)

## Possible Solution
The piece of code that this is pointing at in the contentful.legacy.min.js is ...  function(e=null,t="")
I believe the problem is that IE 11 does not understand default function arguments 
more info: https://stackoverflow.com/questions/44732066/script1006-expected

## Environment

-   **Language Version**: v6.1.0
-   **Package Manager Version**: 3.8.6
-   **Browser Version**: IE 11.914
-   **Operating System**: 64 bit windows

marcolink:
Thanks @kingeric027,

can you please try to pin the version to `7.12.1`:
`<script src='https://cdn.jsdelivr.net/npm/contentful@7.12.1/dist/contentful.legacy.min.js'></script>`

Please let me know if this solves the issue.

PR #390, opened 2020-04-02
fix(build): legacy bundle not IE11 conform

marcolink:
## Summary
With the migration to babel 7 we lost IE11  (es5) support.
These changes fix it and add a `postbuild` validation step.

## Description

The issue was caused by a changed behaviour of env vars inside babel. 
The relevant `env` information is now provided by `api.env()`

A `postbuild` script to check for `es5` compatibility has been added. It runs a compatibility checked on the final bundles. This check failed on the currently released code. 

resolves https://github.com/contentful/contentful.js/issues/371


.babelrc.js
```diff
@@ -69,35 +72,43 @@ var babelConfig = {
...
 
-module.exports = babelConfig
+  if (env === 'test') {
+    babelConfig = Object.assign({}, babelConfig, {
+      presets: [['@babel/preset-env', testBabelPresetEnvConfig]],
+      plugins: babelConfig.plugins.concat(['rewire'])
+    })
+  }
+
+  console.debug('BABEL ENV: ' + env)
```
phoebeschmidt, line 110: do we want these logs in there?

.babelrc.js
```diff
@@ -1,10 +1,9 @@
-var env = process.env.BABEL_ENV || process.env.NODE_ENV
-
 var defaultBabelPresetEnvConfig = {
   // No module transformation, webpack will take care of this if necessary.
   modules: false,
   useBuiltIns: 'entry',
-  corejs: 3
+  corejs: 3,
+  // debug: process.env.NODE_ENV !== 'production'
```
phoebeschmidt, line 6: do we want to keep this in?

.babelrc.js
marcolink: This is a helpful information because it prints the current target (`browser`, `node`, `test`, `legacy`) to the screen. In case of errors, it's obvious which target failed.

.babelrc.js
marcolink: good catch, removed it!

Issue #395, opened 2020-04-11
Unusable with CloudFlare workers due to Axios bug


sgammon:
## Expected Behavior
I would expect to be able to use Contentful JS from a CloudFlare worker.

## Actual Behavior
I cannot use Contentful's Node/JS SDK from a CloudFlare worker because of axios/axios#2507.

## Possible Solution(s)
- Stop using Axios and just use native APIs.
- Switch to a commit of Axios with the proposed fix (axios/axios#2508).
- Write an adapter for Axios to use WebWorker-available tools like `fetch`.

## Steps to Reproduce
1. Create a [CloudFlare Worker](https://developers.cloudflare.com/workers/templates/pages/hello_world).
2. Install Contentful's JS SDK.
3. Follow Contentful's directions to initialize the JS SDK.
4. Observe that, even with minimal code (any code), the following error occurs:

```js
var client = contentful.createClient({
  accessToken: config.accessToken,
  space: config.space
});

// ... later, in a function ...

const response = Contentful.client.getEntries({
    'content_type': 'SomeObject'
});

response.then((data) => {
  console.log('we got some simple data', data);
}, (err) => {
  console.error('we could not fetch some simple data', err);
});
```

<img width="606" alt="Screen Shot 2020-04-11 at 2 08 26 AM" src="https://user-images.githubusercontent.com/171897/79039895-56852300-7b99-11ea-9172-d96efd21004f.png">

## Context
This issue is blocking us from proceeding with Contentful.

## Environment
-   **Language Version**: [CloudFlare Worker](https://developers.cloudflare.com/workers/reference/apis)
-   **Package Version**: `7.14.2`
-   **Which API are you using?**: Delivery (but this issue will affect any network traffic from CFW)

sgammon:
hi contentful devs

people are pissed off about this

you've refused to fix it

we've stopped using your platform because of it

we even pushed code _for you_ to try and fix this, but we just can't maintain it ourselves

there is no reason to use axios anymore when fetch is the norm and XHR is deprecated.

thanks,
a pissed off dev

rande:
I have a quick code to make axios work on `wrangler dev`. This is very the first iteration, and a lot of features are missing, but this can can fire a request to Contentful API.

```
  return createClient({
    space: CONTENTFUL_SPACE_ID as string,
    accessToken: CONTENTFUL_DELIVERY_API as string,
    environment: CONTENTFUL_ENV || "master",
[...]

PR #396, opened 2020-04-12
Fix: Upstream `fetch` adapter for Axios

sgammon:
## Summary
Fix for upstream issues with Axios, which prevent Contentful from working in certain JavaScript contexts (fixes and closes contentful/contentful.js#395).

## Description
Because of the issue reported with contentful/contentful.js#395, caused by upstream issues with Axios (axios/axios#2081, axios/axios#1219, and axios/axios#484 for instance), this change switches the `axios` dependency in `package.json` to a [forked `axios`](axios/axios#2891) that supports the *Fetch API*.

## Motivation and Context
This is necessary to make `axios` (and, therefore, Contentful) work within the parameters of a WebWorker which does not have access to XHR (for instance, a [CloudFlare Worker](https://workers.cloudflare.com/)).

## Todos
- [x] Switched dependency to fork of Axios
- [x] Testing, docs, and so on


sgammon:
the tests seem to pass locally - but CI isn't running. some official help would be great on this

sgammon:
yo... contentful?

phoebeschmidt:
Hi @sgammon, unfortunately in our current configuration, CI can't run for external contributions. We have some plans to fix this but unfortunately not enough time to work on it now.

Even so, this PR isn't mergeable-- we can't change axios to depend on your local fork and we won't remove the logic obscuring auth tokens in responses. 

I agree that Axios not functioning in CloudFlare workers is a problem for the SDK, and I'm sorry it's blocking you. To fix this we'll have to consider different solutions. It's in our backlog and I hope we will get a chance to work on it soon.

sgammon:
@phoebeschmidt okay well the bug is actually that `fetch` is unavailable as an adapter. XHR is obviously old in the first place and carries with it some serious performance implications, in some cases.

fetch avoids these

sgammon:
@phoebeschmidt as prospective customers for Contentful Enterprise we'd want this fixed, and we'd want to not be using XHR if we were using Contentful from the browser, which we have other plans to do. so we'd have to use our fork, and of course that means we have to manually apply and merge any updates that come from upstream

is there really no appetite to write a simple fetch adapter for contentful.js, and avoid use of Axios at all? i don't understand. all we want to do is use your software, and here is a fix - surely, if we can get the upstream PR merged in Axios, we won't have to depend on my fork.

so the real question is, are we willing to work toward fixing this on any reasonable timetable, or do you feel like our being blocked here just isn't important enough to actually get fixed?

sgammon:
@phoebeschmidt if memory serves i removed the token obscuring logic because i couldn't get it to work properly, but there is nothing particularly excluding that logic from working (as far as I know)

with a minimal amount of support from your team we could work together to get that code working, which also eliminates one of the valid concerns you brought up with this PR.

PR #398, opened 2020-04-16
fix(chore): SYS type declaration

marcolink:
## Description
It turned out changes made earlier to `SYS.contentType` are too strict.
Even though, there are rare cases where `contentType` can be missing, we will keep it required for now.

This problem has been reported in context of another [PR](https://github.com/contentful/contentful.js/pull/368) 

**Update:**
We also identified `space` as an optional field and added the changes to this PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -143,7 +143,7 @@ export interface Sys {
     space: {
         sys: SpaceLink;
     };
-    contentType?: {
+    contentType: {
         sys: ContentTypeLink;
     };
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -140,7 +140,7 @@ export interface Sys {
     updatedAt: string;
     locale: string;
     revision?: number;
-    space: {
+    space?: {
         sys: SpaceLink;
     };
     contentType: {


tomimaen:
@marcolink `space` seems to be still a required field, even though in the original PR's description it's said to be left as an optional field to prevent breaking changes.

This is the one change that breaks our build 🙂

marcolink:
> @marcolink `space` seems to be still a required field, even though in the original PR's description it's said to be left as an optional field to prevent breaking changes.
> 
> This is the one change that breaks our build 🙂

@tomimaen 
Thanks for your fast and helpful input on this issue.
 
I changed `space` to be optional . 
Can you confirm this solves the issue you faced in your build process?

tomimaen:
@marcolink Thank you for your fast fix on this one! I can confirm that the build is successful with this change 👍

Aligertor:
Im not sure if somebody cares but i do not like the changes to contentType made here.

The "rare cases" can be produced very reliable imho.
The case i have in mind is:
- Entity A has a linked Entity B
- your editor accidentally deleted Entry B (for what ever reason)
- Entity A now remains with a reference to a non existing Entity
- When querying Entity A now the referenced Entity will have a Sys that has no contentType

This is a case that can happen eventually.
As long as it is possible to delete referenced Entities on https://www.contentful.com/ contentType should be optional.
Imho the developer of the software should be responsible his software handles this error gracefully even so it is a clear mistake on the side of the editor.
Typescript is exactly the tool that should support me in this imho.

PR #423, opened 2020-08-12
Fix/content type optional

phoebeschmidt:
WIP

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -143,7 +143,7 @@ export interface Sys {
     space?: {
         sys: SpaceLink;
     };
-    contentType: {
+    contentType?: {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -133,31 +133,34 @@ export interface SyncCollection {
...
 
-export interface ContentTypeLink {
-    type: 'Link';
-    linkType: 'ContentType';
-    id: string;
+export interface Entity<Type extends EntityType = EntityType> {
+  sys: EntitySys<Type>
+}
+
+export type LinkType = 'Space' | 'Environment' | 'ContentType'
```
janro1, line 157: would be good to have `Entry` here as well

PR #433, opened 2020-10-14
fix(paged-sync): return empty array if sync api returns nothing

### LINKED ISSUE #432: getSyncPage doesnt always return a response
## Expected Behavior

Syncs never crash inside of the Contentful library

## Actual Behavior

We got the following stack trace
```
undefined is not an object (evaluating 't.items') 
main.jsbundle@1004:47215
main.jsbundle@53:155
main.jsbundle@53:864
main.jsbundle@133:581
main.jsbundle@133:830
main.jsbundle@133:2887
main.jsbundle@41:2919
main.jsbundle@41:1095
main.jsbundle@41:2611
main.jsbundle@41:1065
```

Which after analysis with the source map looks like this
```
undefined is not an object (evaluating 't.items')
  at items (./s/contentful/dist/contentful.browser.min.js:1:47663)
  at a (./s/promise/setimmediate/core.js:37:14)
  at cb (./s/promise/setimmediate/core.js:123:25)
  at callback (./s/react-native/Libraries/Core/Timers/JSTimers.js:130:6)
  at passImmediates (./s/react-native/Libraries/Core/Timers/JSTimers.js:181:15)
  at _callImmediatesPass (./s/react-native/Libraries/Core/Timers/JSTimers.js:441:11)
  at _immediatesCallback (./s/react-native/Libraries/BatchedBridge/MessageQueue.js:387:11)
  at __callImmediates (./s/react-native/Libraries/BatchedBridge/MessageQueue.js:135:11)
  at fn (./s/react-native/Libraries/BatchedBridge/MessageQueue.js:364:8)
  at ./s/react-native/Libraries/BatchedBridge/MessageQueue.js:134:17
```

Which is this line [https://github.com/contentful/contentful.js/blob/master/lib/paged-sync.js#L74](https://github.com/contentful/contentful.js/blob/master/lib/paged-sync.js#L74)

And the problem turns out to be in this line [https://github.com/contentful/contentful.js/blob/master/lib/paged-sync.js#L165](https://github.com/contentful/contentful.js/blob/master/lib/paged-sync.js#L165)

## Possible Solution

As per the link just above, the getSyncPage function sometimes (when there is neither `data.nextPageUrl` or `data.nextSyncUrl`) returns nothing at all, which your code clearly doesnt expect, as it then right away tries to access the `items` property of the object it expected to be returned

The solution would be to add an `else` block to getSyncPage so that it always returns an object resembling a response

## Environment

- Environment: React Native 0.63.3
-   **Language Version**: N/A
-   **Package Manager Version**: N/A
-   **Package Version**: 7.14.7

phoebeschmidt:
We were not handling the case where the request to the sync api returned no items, which caused errors in user applications, failing because items was undefined. This PR gracefully handles this case.

Fixes #432

### DIFF
--- a/lib/paged-sync.js
+++ b/lib/paged-sync.js
@@ -146,7 +146,7 @@ function getSyncPage (http, items, query, { paginate }) {
   return http.get('sync', createRequestConfig({ query: query }))
     .then((response) => {
       const data = response.data
-      items = items.concat(data.items)
+      items = items.concat(data.items || [])
       if (data.nextPageUrl) {
         if (paginate) {
           delete query.initial
--- a/lib/paged-sync.js
+++ b/lib/paged-sync.js
@@ -145,7 +145,7 @@ function getSyncPage (http, items, query, { paginate }) {
 
   return http.get('sync', createRequestConfig({ query: query }))
     .then((response) => {
-      const data = response.data
+      const data = response.data || {}
       items = items.concat(data.items || [])
       if (data.nextPageUrl) {
         if (paginate) {
@@ -162,6 +162,8 @@ function getSyncPage (http, items, query, { paginate }) {
           items: items,
           nextSyncToken: getToken(data.nextSyncUrl)
         }
+      } else {
+        return { items: [] }
       }
     })
 }
--- a/test/unit/paged-sync-test.js
+++ b/test/unit/paged-sync-test.js
@@ -43,6 +43,29 @@ test('Throws with incompatible content_type and type parameter', (t) => {
   }, /content_type.*type.*Entry/)
 })
 
+test('Returns empty response if response has no items', (t) => {
+  t.plan(4)
+  const http = { get: sinon.stub() }
+  http.get.withArgs('sync', {
+    params: {
+      initial: true
+    }
+  }).returns(Promise.resolve({
+    data: {
+      nextSyncUrl: 'http://nextsyncurl?sync_token=nextsynctoken'
+    } // no items returned
+  }))
+
+  return pagedSync(http, { initial: true }, { resolveLinks: true })
+    .then(response => {
+      t.deepEqual(response.entries, [])
+      t.deepEqual(response.assets, [])
+      t.deepEqual(response.deletedEntries, [])
+      t.deepEqual(response.deletedAssets, [])
+    })
+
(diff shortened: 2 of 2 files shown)


lib/paged-sync.js
```diff
@@ -146,7 +146,7 @@ function getSyncPage (http, items, query, { paginate }) {
   return http.get('sync', createRequestConfig({ query: query }))
     .then((response) => {
       const data = response.data
-      items = items.concat(data.items)
+      items = items.concat(data.items || [])
       if (data.nextPageUrl) {
```
nfcampos, line 150: I think the issue we are seeing in our application is that nothing is returned from this getSyncPage function, therefore row 74 crashes on trying to access property items of undefined

Which I think would be fixed by adding an else branch in this function, on line 165

marcolink (commented):
@phoebeschmidt great fix ⭐ 

I wonder if this is a breaking change because clients rely on not defined fields?

lib/paged-sync.js
phoebeschmidt: Hi @nfcampos,

thx for clarifying. 

My solution here ensures that `items` will always exist, but you're right that it doesn't cover the case when `response.data` has neither `nextPageUrl` or `nextSyncUrl`. I'm happy to add an extra else here returning an empty `items` array, but would your app expect it to contain a sync token too, or any other properties? Maybe you could tell me a bit more about the case in which the sync api request completes successfully but contains no data (if you know :D )?

phoebeschmidt: I think just an empty items array from `getSyncPage` would work, without any other properties. Then `getSyncPage` will return an empty array. WDYT?

phoebeschmidt:
@marcolink re: breaking change -- 

I actually don't think so since previously the SDK just threw an error because `.items` didn't exist, it never made it to the user as undefined.

lib/paged-sync.js
nfcampos: Yeah I think that makes sense
Sadly I dont have any more details from my app other than what I posted on the issue, not entirely sure what caused it

nfcampos: All looks good for me now after the latest commit:)

Issue #430, opened 2020-09-18
Javascript SDK returning all sys info in items array even if not in select statement


tommaton:
When querying using the select parameter if I search just by the fields I get the fields selected and all the sys object.
but if I query for a item in the sys object I just receive that sys item with no fields

## Expected Behavior

I'm expecting when im doing the following

```
const { items } = await client.getEntries({
    content_type: contentModel.name,
    limit: contentModel.limit,
    select: 'fields.key,fields.value',
  });
```

to only receive the following in the items array

```
"items": [
    {
      "fields": {
        "key": "page.login.passwordErrorMessage",
        "value": "Please enter your password"
      }
    },
    {
      "fields": {
        "key": "page.login.panelLinkText",
        "value": "Set up my online account"
      }
    }
]
```

## Actual Behavior

What I'm actually receiving is:

```
"items": [
    {
      "fields": {
        "key": "page.login.passwordErrorMessage",
        "value": "Please enter your password"
      },
      "sys": {entire sys object}
    },
    {
      "fields": {
        "key": "page.login.panelLinkText",
        "value": "Set up my online account"
      },
      "sys": {entire sys object}
    }
]
```

But if I search for a sys item by id eg
```
const { items } = await client.getEntries({
    content_type: contentModel.name,
    limit: contentModel.limit,
    select: 'sys.id',
  });
```
I get the correct results

```
"items": [
    {
      "sys": {id: 123}
    },
    {
      "sys": {id: 456}
    }
]
```

## Steps to Reproduce
```
const { items } = await client.getEntries({
    // eslint-disable-next-line @typescript-eslint/camelcase
    content_type: 'resource,
    limit: 100,
    select: 'fields.key,fields.value',
  });
```
## Context

This has meant we're receiving larger object responses

## Environment

-   **Language Version**: v12.16.2
-   **Package Manager Version**: yarn 1.22.5
-   **Operating System**: Mac OS
-   **Package Version**: 7.14.6

Issue #296, opened 2018-12-19
Link Resolution doesn't work if select contains sys.id


erik-wendt:
## Expected Behavior

I'm using the JavaScript SDK and try to "select" only some fields from the content entries and the ID of it, e.g. "select: 'sys.id,fields.route,fields.pages". 

The "field.pages" are links to other content entries and with param "include: 3" I expect to get the result with link resolution enabled. 

## Actual Behavior

If I'm Using "sys.id" in the "select" param the link resolution doesn't work anymore and I get the data without the related fields.

## Possible Solution

workaround: remove "sys.id" from the select param

## Example

I have a simple content model "page" which contains a field "route" and a field "pages" for linked entries of the same model. 

```
client
        .getEntries({
            content_type: 'page',
            // link resolution works! will contain all sys data
            //select: 'fields.route,fields.pages',
            // link resolution doesn't work. will contain sys.id only
            select: 'sys.id,fields.route,fields.pages',
            include: 3
        })
        .then((result) => result.items)
        .then((pages) => {
            pages.forEach((page) => console.log(page.sys, page.fields));
        })
        .catch(console.error)
```

## Environment

-   **Browser Version**: Chrome 70.0.3538.110
-   **Operating System**: macOS Serria
-   **Package Version**: v7.0.5
-   **Which API are you using?**: Delivery API

erik-wendt:
BTW: it worked before SDK Version 5.1.0. If I'm using 5.0.5 I get the expected result.

marcolink:
Hi @erik-wendt,

could you provide an actual and expected example result?

erik-wendt:
Hi @marcolink,

I've tested it again for the new version "7.14.2" and it's still not working.

If I use this query the link resolution just works as expected

```
client.getEntries({
  content_type: 'page',
  // link resolution works! will contain all sys data
  select: 'fields.route,fields.pages',
  include: 3
})
``` 
and produces the following output:

![image](https://user-images.githubusercontent.com/7628803/78771393-61b93280-7990-11ea-990c-20dd1d226772.png)


But if i try to filter the "sys" object to get the ID only the link resolution is not working anymore
```
client.getEntries({
  content_type: 'page',
  // link resolution doesn't work. will contain sys.id only
  select: 'sys.id,fields.route,fields.pages',
  include: 3
})
```
![image](https://user-images.githubusercontent.com/7628803/78771352-536b1680-7990-11ea-977f-2ba0c1fe4a28.png)

[...]

Issue #444, opened 2020-11-04
The removeUnresolved flag unexpected behaviour


aelmeniawy:
Our application consumes Contentful with linked draft entries. We went for using the flag provided by the SDK `removeUnresolved` to filter out the problematic draft entries but it seemed to filter out more than what we wanted.

Sometimes we fetch entries using a certain depth in order just to fetch the sys property of the entry and not include the fields.

## Expected Behavior

When the flag `removeUnresolved` is set to true it would remove only draft entries from the response object..

## Actual Behavior

Entries that don't contain the `.fields` property get filtered out even though they're not in draft mode.

## Possible Solution

What would be nice to have is to only filter out entries that are actually in draft mode and not filter out the entries that are fetched with shallow depth (i.e., not including the .fields property).

## Steps to Reproduce

1. Write a query that fetches only the sys property of entries linked within an entry
2. Set the `removeUnresolved` flag to `true`

Entries within the entry fetched would all be filtered out as they don't contain the `.fields` property.

## Context

I was trying to filter out draft entries.

## Environment

-   **Language Version**:`12.x`
-   **Package Version**:  v7.14.8 
-   **Which API are you using?**:  Delivery

TimBeyer:
Hi @aelmeniawy, I'm wondering if you encountered this bug when you did a `select: 'sys'` or if you actually selected a specific set of `sys` properties. In theory the filter should really only remove entities it cannot find and the `fields` property is not looked at. There's a different bug however that sometimes makes the SDK not resolve links correctly in case a subset of `sys` fields was used. I'm wondering if this might be another instance of that bug or if I need to dig a bit deeper here.

TimBeyer:
@aelmeniawy could you check if the latest release also fixed your problem?

nenad-percic:
@TimBeyer @aelmeniawy is colleague of mine on the same project - version 7.15.1 is currently not testable for us due to #503 that we reported today. I will followup this as soon as it becomes actionable.

Thank you!

TimBeyer:
Your use case should now work in https://github.com/contentful/contentful.js/releases/tag/v8.1.0

nenad-percic:
@TimBeyer That was unexpectedly quick release. :)

I wrote two [tests](https://prnt.sc/vxgzlx) that would cover our use-case and both of them are green! :) 
Thank you very much!

PR #492, opened 2020-11-30
fix: ensure minimum select

TimBeyer:
Fixes https://github.com/contentful/contentful.js/issues/296 and sort of fixes https://github.com/contentful/contentful.js/issues/430
Might be the cause for https://github.com/contentful/contentful.js/issues/444 too.

Basically the SDK always needs `sys.id` and `sys.type` to perform link resolution.  
Other `sys` fields are optional from what I can tell.

Previously the SDK would simply check if the string `sys` appears in the `select` section of the query and continue with that if it did, and add `sys` if it did not.

This led to the problem where selecting for example just `sys.id` would fail link resolution because `sys.type` was not included. It also led to the problem where not selecting `sys` as an optimization would always force the entire sys object back into the query. 

Now we ensure that the minimum set of sys properties is always included, but not more than that.
This does not fully fix the issue of not wanting `sys` at all, but it gives the best compromise for still keeping things working.

### DIFF
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -48,6 +48,7 @@
 import { createRequestConfig } from 'contentful-sdk-core'
 import entities from './entities'
 import pagedSync from './paged-sync'
+import normalizeSelect from './utils/normalize-select'
 
 /**
  * Creates API object with methods to access functionality from Contentful's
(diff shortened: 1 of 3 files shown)


matthew-contentful (approved):
LGTM

PR #496, opened 2020-12-02
refactor: replace promises with async/await

TimBeyer:
BREAKING CHANGE: Some functions that used to throw synchronously for things like bad parameters now reject the promise.
In many cases users may not have to do anything assuming the calls already happened within a promise chain, but in rare cases this may need some refactoring of error handling cases.

### DIFF
--- a/lib/create-contentful-api.js
+++ b/lib/create-contentful-api.js
@@ -105,14 +105,17 @@ export default function createContentfulApi ({ http, getGlobalOptions }) {
    *   accessToken: '<content_delivery_api_key>'
    * })
    * // returns the space object with the above <space-id>
-   * client.getSpace()
-   * .then((space) => console.log(space))
-   * .catch(console.error)
+   * const space = await client.getSpace()
+   * console.log(space)
    */
-  function getSpace () {
+  async function getSpace () {
     switchToSpace(http)
-    return http.get('')
-      .then((response) => wrapSpace(response.data), errorHandler)
+    try {
+      const response = await http.get('')
+      return wrapSpace(response.data)
+    } catch (error) {
+      errorHandler(error)
+    }
   }
 
   /**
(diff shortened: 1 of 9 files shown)


marcolink (approved):
I think the PR `title` should start with `BREAKING CHANGE:`  to trigger a major release
https://github.com/semantic-release/semantic-release#commit-message-format

TimBeyer:
According to https://www.conventionalcommits.org/en/v1.0.0/r the breaking change section goes into the footer or you add the exclamation mark after the type.

suevalov (approved):

mduleone:
Hello contentful team! Question about this change: Previously, we were able to run contentful in Node (v12.3.1) without issue, and now the same script is throwing an error. I understand that this includes breaking changes, but what would you recommend as the course of action to resolve something like this?
```sh
(node:72864) UnhandledPromiseRejectionWarning: ReferenceError: regeneratorRuntime is not defined
    at Object._getEntries (/Users/mattduleone/sites/next-web/node_modules/contentful/dist/contentful.node.js:6213:51)
    at Object.getEntries (/Users/mattduleone/sites/next-web/node_modules/contentful/dist/contentful.node.js:6191:24)
    at fetchContentfulData (/Users/mattduleone/sites/next-web/bin/contentful-refresh.js:59:33)
    at populateData (/Users/mattduleone/sites/next-web/bin/contentful-refresh.js:90:29)
    at Object.<anonymous> (/Users/mattduleone/sites/next-web/bin/contentful-refresh.js:111:1)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:785:10)
    at Module.load (internal/modules/cjs/loader.js:641:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:837:10)
```

TimBeyer:
Oh that was definitely an unintended side effect of this change.  
For now to unblock you, you could use https://www.npmjs.com/package/regenerator-runtime directly in your codebase.  
For a proper fix I think we will bump the minimum node version required to run this because it seems a bit much to make everyone on a modern version of node use regenerator when all of this is natively supported.

TimBeyer:
@mduleone we have a fix prepared in https://github.com/contentful/contentful.js/pull/498 that bumps the minimum version of node and thus gets rid of the dependency for regenerator.

mduleone:
@TimBeyer thank you! Testing this out now!

mduleone:
@TimBeyer Everything is working as expected. Thank you again!

jesperp:
Hi, also getting the same error as @mduleone (`ReferenceError: regeneratorRuntime is not defined`), but inside the browser! (using Svelte, Typescript, Snowpack)

I'm just passed fixing this error:
https://github.com/contentful/contentful-sdk-core/issues/120

And now trying to get a simple `client.getEntries` call working but getting the regeneratorRuntime error. Installing the regenerator-runtime package and importing it fixes this but can this be solved by contentful? I'm a bit javascript fatigued after many strange errors leading way down the rabbit hole of SO posts and github issues. Maybe the error is with svelte preprocessing / snowpack, I don't know! Hope to get some light on this issue anyhow!

Issue #503, opened 2020-12-07
TypeError: query.select.split is not a function


jugglingthebits:
My team is seeing failures in the Contentful client since the update from 7.14.13 to 7.15.2.

## Actual Behavior

We're having numerous tests fail with the following error:

```
TypeError: query.select.split is not a function
      at normalizeSelect (node_modules/contentful/dist/webpack:/contentful/utils/normalize-select.js:14:35)
      at Object.getEntries (node_modules/contentful/dist/webpack:/contentful/create-contentful-api.js:216:20)
      at getCachedContentfulEntriesOrFetch (src/clients/contentful/contentful-client.ts:616:44)`
```

FYI Updating to 8.0.3 exhibits the same error
```
TypeError: query.select.split is not a function
    at normalizeSelect (/home/circleci/alma/api/node_modules/contentful/dist/webpack:/contentful/utils/normalize-select.js:14:35)
    at Object.getEntries (/home/circleci/alma/api/node_modules/contentful/dist/webpack:/contentful/create-contentful-api.js:227:20)
    at getCachedContentfulEntriesOrFetch (/home/circleci/alma/api/src/clients/contentful/contentful-client.ts:616:44)
```

## Context

The client is breaking production code.

## Environment

-   **Language Version**: 12.20.0
-   **Operating System**: Docker image `cimg/node:12.20.0`
-   **Package Version**: 7.15.2
-   **Which API are you using?**: Delivery

TimBeyer:
Could you share what you are passing as `query.select`?
We [fixed a bug with the select behavior in 7.15.1](https://github.com/contentful/contentful.js/pull/492) with the existing code indicating that the `select` option will always be a string.  
Of course the querystring serializer might still take an array of strings there, and it seems we simply use the default [`qs.stringify`](https://github.com/contentful/contentful-sdk-core/blob/master/src/create-http-client.ts#L98) behavior there which as a default would be using the array notation like `foo[0]=a&foo[1]=b&foo[2]=c` which isn't supported in the API according to [the docs](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/select-operator).  
So I'm wondering if you were calling it with an array of strings, and if that actually did show the expected behavior pointing to either a different serializer or the API secretly parsing that kind of syntax as well.

jugglingthebits:
@TimBeyer We observe numerous different queries failing. Here is an example query (passed to `getEntries<T>(query?: any)`:
[...]

PR #504, opened 2020-12-07
feat: allow arrays for select

TimBeyer:
Previously we did not officially, support arrays for the select operator, though it would sometimes work nonetheless.
This change ensures that we do support arrays and that they too are normalized properly.
Additionally, we stop mutating the query object in the select normalization and always return a copy.

Fixes https://github.com/contentful/contentful.js/issues/503


marcolink (approved):
LGTM

PR #511, opened 2020-12-08
Remove entities

marcolink:
- Remove all entities to avoid unnecessary cloning of result entries. 
- Add types to contentful-api


lib/utils/resolve-circular.ts
```diff
@@ -0,0 +1,11 @@
+import copy from 'fast-copy'
+import mixinStringifySafe from '../mixins/stringify-safe'
+import resolveResponse from 'contentful-resolve-response'
+
+export function resolveCircular (data, { resolveLinks, removeUnresolved }) {
+  const wrappedData = mixinStringifySafe(copy(data))
```
TimBeyer, line 6: Actually `resolveResponse` [already makes a copy](https://github.com/contentful/contentful-resolve-response/blob/master/index.js#L114) so I think we can skip that copy as well.

marcolink: @TimBeyer yay, that was the last ref to `fast-copy` :)

TimBeyer (approved):

PR #513, opened 2020-12-08
feat(error-handling): introduce not found error

marcolink:


lib/create-contentful-api.ts
```diff
@@ -279,352 +306,360 @@ interface GetConfig {
...
+     *   space: '<space_id>',
+     *   accessToken: '<content_delivery_api_key>'
+     * })
+     *
+     * const entry = await client.getEntry('<entry_id>')
+     * console.log(entry)
+     */
+    async function getEntry<T>(id, query = {}): Promise<Entry<T>> {
+        if (!id) {
+            throw notFoundError(id)
```
TimBeyer, line 448: Seems a bit weird to throw a not found error that we pass the ID that doesn't exist :)

lib/create-contentful-api.ts
```diff
@@ -279,352 +306,360 @@ interface GetConfig {
...
+            id,
+            environment,
+            space
+        }
+    }
+}
+
+export default function createContentfulApi({http, getGlobalOptions}): ContentfulClientApi {
+    const notFoundError = (id) => {
+        return new NotFoundError(id, getGlobalOptions().environment, getGlobalOptions().space)
```
TimBeyer, line 331: Maybe we should call the `getGlobalOptions` function before and ensure that this is all the same everywhere?  
The mix of the `switchToSpace` side effects and calling the suspiciously named `getGlobalOptions` function gives me a bit of anxiety 😉

lib/create-contentful-api.ts
marcolink: @TimBeyer yeah, but that part is not new :)

PR #514, opened 2020-12-08
chore(lint): use tsdx defaults

marcolink:


marcolink:
We can close this in favor of #527

PR #516, opened 2020-12-09
feat(tests): migrate all create-contenful-api tests

marcolink:


test/unit/create-contentful-api-parseentries.test.ts
```diff
@@ -0,0 +1,61 @@
+import createContentfulApi from '../../lib/create-contentful-api'
+import createGlobalOptions from '../../lib/create-global-options'
+
+test('Given json should be parsed correctly as a collection of entries', () => {
```
marcolink, line 4: I had to extract this test into a single file to avoid a complex setup with mocking.
The test is still failing, which might or might not be due to the changes we did.

marcolink: Ok, it's passing now, the global configuration was wrongly setup

TimBeyer (approved):

PR #517, opened 2020-12-09
feat(types): typed public api

marcolink:
Most of the public API typed.

### DIFF
--- a/lib/common-types.ts
+++ b/lib/common-types.ts
@@ -1,0 +1,175 @@
+import { AxiosInstance } from 'contentful-sdk-core'
+
+export type HttpClientInstance = AxiosInstance
+export interface Asset {
+  sys: Sys;
+  fields: {
+    title: string;
+    description: string;
+    file: {
+      url: string;
+      details: {
+        size: number;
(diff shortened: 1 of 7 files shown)


lib/common-types.ts
```diff
@@ -0,0 +1,175 @@
...
+  type Link<T> = Asset | Entry<T>;
+  type Array<T = any> = Symbol[] | Entry<T>[] | Asset[];
+  type Object<T = any> = T;
+  interface RichText {
+    data: {};
+    content: RichTextContent[];
+    nodeType: 'document';
+  }
+}
+interface RichTextDataTarget {
```
TimBeyer, line 150: Shall we already import the rich text types from the separate project here?

TimBeyer:
There's some changes I think are required for a full release but I'd say for now that's a step forward already

TimBeyer (approved):

lib/common-types.ts
marcolink: @TimBeyer I think it's worth a try, but we should do it in a separate PR

PR #531, opened 2020-12-14
feat(build): output right files

marcolink:
output the same set of files as we do with the old version

Screenshot 2020-12-14 at 17 58 56


Vinz93 (approved):
LGTM 👍

Issue #135, opened 2017-03-31
Is there a way to exclude certain fields from a `getEntries` call?


mrmartineau:
* **Node Version**: v7.7.2
* **Npm Version**: 4.4.4
* **Platform**: macOS Sierra
* **Package Version**: 4.1.2

If I have a list of articles, is it possible to exclude the article body from a `client.getEntries` call? I would like to do this to speed up the response from the call when listing all the articles. Does that make sense?

Khaledgarbaya:
Hi @mrmartineau Yes you can do that using the select operator.
```js
client.getEntries({content_tytpe: 'myContentTypeId', select: 'fields.fieldName, fields.fieldName2'})
```
For more infos on that check the [API documentation](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/select-operator).
One limitation when using the SDK the `sys` property will be appended always since it is needed fo the link resolution.
**Hint:** we added SDK snippets in the documentation, so can copy paste code and run it very easily just click on the drop down and choose Javascript.

mrmartineau:
Hi @Khaledgarbaya, I actually just found that in the docs and an example while [searching your repo](https://github.com/contentful/contentful.js/blob/master/test/integration/tests.js#L100-L109), thank you 👍 

I have to say that while your API is brilliant, I find it hard to navigate the various docs because I am never quite sure which one is the right one. I am using the [js sdk docs](https://contentful.github.io/contentful.js/contentful/4.1.2/), but it doesn't have documentation for all aspects of the API, so then I try the readme or the [API docs](https://www.contentful.com/developers/docs/references/content-delivery-api/). I'm not sure how it should be fixed, or even if it needs to be fixed, but I thought I should highlight a small issue I have with them..

Khaledgarbaya:
Hi @mrmartineau it is going to be very hard to cover every aspect of the API in the SDK documentation, we are working on Adding more and more examples to the documentations, the new API docs already containing tons of snippets I would encourage you to search there for now, but feel free to open any issue anytime, I will be glad to help you.

Best,
Khaled

mrmartineau:
Thanks @Khaledgarbaya, as I said I'm not sure if you do need to change it, but I certainly found it a little confusing as to where I should look for something..

jellevandevelde:
The Content Delivery API docs are a nightmare to navigate I find. When selecting fields, make sure to only separete them using commas, without spaces.

As such:
[...]

PR #534, opened 2020-12-16
Ignore spaces around commas in select parameter

aatishnn:
## Summary
Few people including myself have run into issues when trying to use the `select` parameter in queries with spaces around fields separated by commas.  This simple PR fixes that by trimming the spaces around.

## Motivation and Context
People have run into same issue as I did:
https://github.com/contentful/contentful.js/issues/135#issuecomment-490431849


marcolink (approved):
@aatishnn thanks for the fix 🎉  - looks good!

PR #556, opened 2021-01-13
Dedicated `sys` types

marcolink:
## Summary
Dedicated types for different entities.

## Description
Given the different shapes of the `sys` field between the different entities, we want to reflect that also on a type level.


ruderngespra (approved):
LGTM! Just one follow up question: The `baseSys` here has a `locale` property, but tags don't have this. Wouldn't this become a bit annoying in the future?

marcolink:
> LGTM! Just one follow up question: The `baseSys` here has a `locale` property, but tags don't have this. Wouldn't this become a bit annoying in the future?
> 
> But probably we should not worry about this at this early stage with no tags available in the cda?

@ruderngespra You're right!
Please check the reshaped types with a very basic `BaseSys`

damienxy (approved):

lib/common-types.ts
```diff
@@ -1,91 +1,126 @@
...
   revision?: number
-  space?: {
-    sys: SpaceLink
-  }
-  contentType: {
-    sys: ContentTypeLink
-  }
+  space?: { sys: SpaceLink }
+  environment: { sys: EnvironmentLink }
+  locale: string
```
TimBeyer, line 90: I think `locale` is only present when you don't request via `locale=*`.
So probably we will have to split this out as well.

lib/common-types.ts
```diff
@@ -1,91 +1,126 @@
...
   updatedAt: string
-  locale: string
   revision?: number
-  space?: {
-    sys: SpaceLink
-  }
-  contentType: {
-    sys: ContentTypeLink
-  }
+  space?: { sys: SpaceLink }
```
TimBeyer, line 88: Do we know when `space` is actually not present?  
Those optional properties will always force people to check the existence in their code.
Maybe we already know if it exists or not?

lib/common-types.ts
```diff
@@ -1,91 +1,126 @@
...
+interface BaseSys {
   type: string
   id: string
+}
+
+type EntitySys = {
   createdAt: string
   updatedAt: string
-  locale: string
   revision?: number
```
TimBeyer, line 87: When is `revision` not present?
I think for entries, assets and content types it's always there.

lib/common-types.ts
```diff
@@ -1,91 +1,126 @@
...
+export interface Sys extends EntitySys {
+  contentType: { sys: ContentTypeLink }
 }
-export interface Link<T extends string> {
+
+export type AssetSys = EntitySys
+
+export type SpaceSys = { type: 'Space' } & BaseSys
+
+export type ContentTypeSys = {
```
TimBeyer, line 101: The pattern in the preceding types was to use `interface Foo extends Bar`.
Maybe we should do that here in the new sys types  as well for consistency.

PR #557, opened 2021-01-13
Use rich text types package

marcolink:
## Description

The locally defined types for rich text content are inaccurate. 
Instead, we use a dedicated types package for `rich-text- types`.


ruderngespra (approved):
:+1:

damienxy (approved):

PR #563, opened 2021-01-14
TS Type support for queries

marcolink:
## Summary
Typed `query` parameter for `getEntries` and `getAssets` as part of the typescript migration initiative.

## Description
New types for queries to help writing the right query matching your `Entry` shape. 
We use template string literals to create all relevant/possible keys based on the shape of the entry type you want to query. 
Currently, we don't do strict type checking, but rather expect a significant better editor experience. 

### Example `fields`
```typescript
type ExampleFields = {
  name: string,
  collection: number[]
  dateField: EntryFields.Date
  center: EntryFields.Location
}

const entries = await client.getEntries<ExampleFields>({
  'fields.name' : 'hello world',
  'fields.center[near]' : [0, 0, 100],
});
````
> all possible field strings will be suggested with all possible filters

### Example `select`
```typescript
type ExampleFields = {
  name: string,
  collection: number[]
  dateField: EntryFields.Date
  center: EntryFields.Location
}

const entries = await client.getEntries<ExampleFields>({
  select:['fields.center', 'sys.space']
});
````
> all possible strings will be suggested by your editor

### Limitations
- we only take the root level fields into account to not exhaust the type system.

## Todos
-   [ ] add exmaple and docs to README

## Screenshots:
Screenshot 2021-01-14 at 21 52 42

### CODE REVIEW

File: lib/create-contentful-api.ts [hunk 1]
```diff
@@ -14,14 +14,13 @@ import {
   ContentType,
   ContentTypeCollection,
   Entry,
-  EntryCollection,
+  EntryCollection, EntryQueries,
```
damienxy, line 17: would put this in a new line

lib/types/entry.ts
```diff
@@ -16,7 +16,7 @@ export declare namespace EntryFields {
   type Text = string
   type Integer = number
   type Number = number
-  type Date = string
+  type Date = `${number}-${number}-${number}T${number}:${number}:${number}Z`
```
damienxy, line 19: is that the only date format that can ever happen?

lib/types/query/fts.ts
```diff
@@ -0,0 +1,16 @@
...
+// In the future figure out how to really exclude the `Date` type
+type SupportedTypes = Exclude<(EntryFields.Symbol | EntryFields.Text | EntryFields.RichText), EntryFields.Date>
+
+export type FullTextSearchFilters<Fields, Prefix extends string> = {
+  [FieldName in keyof ConditionalPick<Fields, SupportedTypes> as `${Prefix}.${string & FieldName}[${FilterType}]`]?:
+  BaseOrArrayType<Fields[FieldName]> extends EntryFields.Date
+    ? never
+    : BaseOrArrayType<Fields[FieldName]> extends SupportedTypes
+    ? string
+    : never
```
damienxy, line 15: can this be flipped to use `never` only once?

test/unit/mocks.ts
```diff
@@ -7,8 +7,11 @@ import {
   EnvironmentLink,
   Locale,
   EntrySys,
-  SpaceLink,
+  SpaceLink, EntryFields
```
damienxy, line 10: would put this in a new line

test/unit/types.test.ts
```diff
@@ -1,15 +1,17 @@
 import {
   Entry,
-  FieldsQueries,
+  EntryQueries,
   EntrySys,
   LocaleSys,
   ContentTypeSys,
   AssetSys,
-  SpaceSys,
+  SpaceSys, EntryFields
```
damienxy, line 8: would put this in a new line

damienxy (commented):
Really nice!
I would approve but it wouldn't mean anything cause I don't know our data structures enough :-)

lib/create-contentful-api.ts
```diff
@@ -30,19 +29,19 @@ export interface ContentfulClientApi {
 
   getAsset(id: string): Promise<Asset>
 
-  getAssets(query?: FieldsQueries<AssetFields>): Promise<AssetCollection>
+  getAssets(query?: EntryQueries<AssetFields>): Promise<AssetCollection>
```
TimBeyer, line 32: Maybe we should alias `AssetQueries` somewhere else and pass that in directly here.  
It looks sort of weird to see `EntryQueries` in here :)

lib/types/query/query.ts
```diff
@@ -0,0 +1,62 @@
...
+  SelectQueries<Fields, 'fields'>
+
+export type EntryQueries<Fields extends FieldsType = FieldsType> =
+  EntryFieldsQueries<Fields> &
+  SysQueries<EntrySys> &
+  FixedQueryOptions &
+  FixedPagedOptions &
+  Record<string, any>
+
+function query<Fields>(query: EntryQueries<Fields>): void {
```
TimBeyer, line 47: I guess this and line 59 onwards need to still be removed.

lib/create-contentful-api.ts
marcolink: @TimBeyer good point - done!

lib/types/query/query.ts
marcolink: @TimBeyer done!

lib/create-contentful-api.ts
```diff
@@ -30,19 +30,19 @@ export interface ContentfulClientApi {
...
 
   getEntries<Fields = Record<string, any>>(
-    query?: FieldsQueries<Fields>
+    query?: EntryQueries<Fields>
   ): Promise<EntryCollection<Fields>>
 
   getEntry<Fields = Record<string, any>>(
     id: string,
-    query?: FieldsQueries<Fields>
+    query?: EntryQueries<Fields>
```
TimBeyer, line 45: I am still confused what kind of query we would send together with fetching an entry by ID.
Things like `select`?

TimBeyer (approved):

lib/types/entry.ts
marcolink: Yes!

PR #594, opened 2021-02-08
feat(tags): adjust unit test mocks to expect metadata on entities

ruderngespra:


marcolink:
> Updated the space used for integration tests to have private and public tags attached to an entry and to have an entry with a field named metadata

@ruderngespra thanks that you already set up a test for this 👏 

Would it be possible to test the `fields` contains a field named `metadata` scenario with a mocked object as well? 
It could be a good example for other SDKs.

marcolink:
@ruderngespra added more tests, I think we're good to go, right?

marcolink (approved):

PR #625, opened 2021-02-23
fix(chore): bundle building

marcolink:
## Summary
Fixes the bundling of this library.

### Observation

When doing a webpack build with a simple setup like [this](https://github.com/rustyy/contentful-issue-545), you get a `ReferenceError`:

```
Uncaught ReferenceError: regeneratorRuntime is not defined
```

I managed to reproduce the problem. As a quick fix for now, I suggest you replace:

```javascript
import {createClient} from 'contentful';
// and same for
import {createClient} from 'contentful/dist/contentful.legacy';
```

with:

```javascript
import {createClient} from 'contentful/dist/contentful.browser';
```
> the browser bundle has native async support

**The reason why this is happening is that some bundles expect the regeneratorRuntime, but we don't bundle them.**

### Currently exported bundles:

- `contentful` points to `dist/es-modules/contentful.js` (expects regeneratorRuntime)
- `contentful.browser` points to `dist/contentful.browser.js` (uses native async, due to the full support of the feature based on our current [browser-list](https://github.com/contentful/browserslist-config/blob/master/index.js))
- `contentful.legacy` points to points to `dist/contentful.legacy.js` (expects regeneratorRuntime, due to the manually defined browser list: `['last 5 versions', 'not ie < 10']`)
- `contentful.node` points to `dist/contentful.node.js` (expects regeneratorRuntime)

## Solution

Things to keep in mind:
> `contentful.browser` and `contentful.legacy` are both used via [jsDeliver](https://www.jsdelivr.com/package/npm/contentful), directly embedded in your HTML file. Without having `regenerator-runtime` bundled, they wont work out of the box as before (< 8). 

### A good set of bundles that I can think of would be:

- default -> es6 modules with native async
- legacy & using bundled regeneratorRuntime
- node & browser -> basically a bundled single file version of es6 modules with native async

## ToDo:

- [x] `node` build as bundled umd file without generators
- [x] `browser` build as bundled umd file without generators
- [x] `legacy` build as bundled umd file without generators
- [x] `es6` with native async

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -52,7 +52,7 @@ const defaultBabelLoader = {
   options: {}
 }
 
-// Browsers
+// Browsers with native async await calls
 const browserBundle = copy(baseBundleConfig)
 browserBundle.module.rules = [
   Object.assign({}, defaultBabelLoader, {
@@ -63,7 +63,7 @@ browserBundle.module.rules = [
 ]
 browserBundle.output.filename = `${baseFileName}.browser${PROD ? '.min' : ''}.js`
 
-// Legacy browsers like IE11
+// Legacy browsers like IE11, with bundled regenerator-runtime
 const legacyBundle = copy(baseBundleConfig)
 legacyBundle.module.rules = [
   Object.assign({}, defaultBabelLoader, {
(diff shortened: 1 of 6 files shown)


webpack.config.js
```diff
@@ -52,7 +52,7 @@ const defaultBabelLoader = {
   options: {}
 }
 
-// Browsers
+// Node bundled umd file, with native async await calls
 const browserBundle = copy(baseBundleConfig)
```
TimBeyer, line 56: If this is the node bundle, shall we rename this variable to not add extra confusion?

marcolink: @TimBeyer Good catch - fixed it!

marcolink: btw - it's the browser bundle, must have mixed it up when writing that comment

TimBeyer (approved):

PR #648, opened 2021-03-04
OutgoingMessage.prototype._headers deprecated

rrostt:
## Summary

Removes deprecated code. Cannot replace with setHeaders since headers already sent.
In case you still want to obscure the headers from the request-object i suggest you strip down the request object to only include desired attributes, rather than changing the original object.

## Description

OutgoingMessage.prototype._headers is deprecated and use results in error.

https://nodejs.org/api/deprecations.html#DEP0066

## Motivation and Context

Fix needed cos code crashes on node 15.

## Todos

-   [ ] Implemented feature


jesseday:
Is there any movement on this PR?  I'm getting this message during builds.

I'm on Node 12.22.4

woodbrettm:
Same thing happening for me

makinwab:
Thanks for helping with this, @rrostt.

PR #732, opened 2021-04-30
fix: add new content model to tests

mshaaban0:
## Summary

We added a new content model to test entry with multiple [references](https://github.com/contentful/contentful-management.js/pull/829), this causes tests to fail in other PRs

## Checklist

- [x] Both unit and integration tests are passing
- [x] There are no breaking changes


marcolink (approved):
:shipit:

PR #807, opened 2021-06-16
fix: add tag type definitions

### LINKED ISSUE #806: Typescript: types do not include new tags api methods in 8.4.1
## Expected Behavior
`getTag()` and `getTags()` functions exposed on `ContentfulClientApi` interface to match updated implementation.

## Actual Behavior
Functions missing ...

```
export interface ContentfulClientApi {
    getAsset(id: string, query?: any): Promise<Asset>;
    getAssets(query?: any): Promise<AssetCollection>;
    getContentType(id: string): Promise<ContentType>;
    getContentTypes(query?: any): Promise<ContentTypeCollection>;
    getEntries<T>(query?: any): Promise<EntryCollection<T>>;
    getEntry<T>(id: string, query?: any): Promise<Entry<T>>;
    getSpace(): Promise<Space>;
    getLocales(): Promise<LocaleCollection>;
    parseEntries<T>(raw: any): Promise<EntryCollection<T>>;
    sync(query: any): Promise<SyncCollection>;
}
```

## Possible Solution
Update types

## Steps to Reproduce
Try to code-complete with typescript or view `index.d.ts`

## Context
Cannot use new tag functionality without `@ts-ignore`

## Environment
-   **Language Version**: irrelevant, but v14.16.1
-   **Package Manager Version**: irrelevant, but 7.10.0
-   **Browser Version**: node
-   **Operating System**:  macOS
-   **Package Version**: 8.4.1
-   **Which API are you using?**: `ContentfulClientApi`

anho:
## Summary

Includes missing tag types, fixes #806.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -133,7 +133,7 @@ export interface Tag {
       id: string
       type: 'Tag'
       version: number
-      visibility: string
+      visibility: 'public'
     }
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,6 +43,8 @@ export interface ContentfulClientApi {
     getEntry<T>(id: string, query?: any): Promise<Entry<T>>;
     getSpace(): Promise<Space>;
     getLocales(): Promise<LocaleCollection>;
+    getTag(id: string): Promise<Tag>;
+    getTags(query?: any): Promise<TagCollection>;
     parseEntries<T>(raw: any): Promise<EntryCollection<T>>;
     sync(query: any): Promise<SyncCollection>;
 }
@@ -124,6 +126,18 @@ export interface Locale {
 }
 
 export type LocaleCollection = ContentfulCollection<Locale>;
+
+export interface Tag {
+    name: string
+    sys: {
+      id: string
+      type: 'Tag'
+      version: number
+      visibility: string
+    }
+}
+
+export type TagCollection = ContentfulCollection<Tag>;
 
 export interface SyncCollection {
     entries: Array<Entry<any>>;


index.d.ts
```diff
@@ -125,6 +127,18 @@ export interface Locale {
 
 export type LocaleCollection = ContentfulCollection<Locale>;
 
+export interface Tag {
+    name: string
+    sys: {
+      id: string
+      type: 'Tag'
+      version: number
+      visibility: string
```
makinwab, line 136: Since visibility can only be one of `public` or `private`, better to explicitly use those instead like so `'public' | 'private'`?

marcolink: Maybe just make it `public` - since we only deliver those on the CDA?

anho, line 136:
```suggestion
      visibility: 'public'
```

makinwab (approved):

index.d.ts
```diff
@@ -125,6 +127,18 @@ export interface Locale {
 
 export type LocaleCollection = ContentfulCollection<Locale>;
 
+export interface Tag {
+    name: string
+    sys: {
```
marcolink, line 132: Should we define the full received `sys` object here?
Reference: https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/content-tags/tag/get-a-single-tag/console/curl

anho: It's not done on the other entities, so I'd rather spare the effort of putting that information in. I heard someone was working on some TS refactoring ;)

PR #854, opened 2021-08-18
feat(chore): setup dependabot

marcolink:
setup GH native dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+version: 2
+registries:
+  npm-registry-registry-npmjs-org:
+    type: npm-registry
+    url: https://registry.npmjs.org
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,24 @@
...
+updates:
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "00:00"
+      timezone: UTC
+    open-pull-requests-limit: 10
+    ignore:
+    - dependency-name: husky
```
damienxy, line 17: I think this is no longer needed

marcolink: Good point, let's open a separate PR for that!

marcolink:
@damienxy I added another action to auto-merge only dependabot pull requests. With this we don't have to activate auto-merge on all pull requests.

damienxy (approved):

PR #863, opened 2021-08-31
fix(dependabot): do not refer to npm registry in config

ruderngespra:
Remove registry references as we don't rely on private dependencies and the reference broke our setup.


marcolink (approved):
ah, that's a learning 🙇‍♂️

PR #899, opened 2021-09-13
feat(errors): improve error data thrown

makinwab:
BREAKING CHANGE: stop sending Axios error object

## Summary
Introduce a new error handling function and parse Axios' error object.

## Description
Use the error handling function from the contentful-sdk-core library that handles and returns better-parsed errors.

## Motivation and Context
We want to obscure tokens returned in the error response object. With this change, we're implementing a better way to obfuscate and parse the error response we get from Axios.

## Todos

-   [x] Implemented feature

### DIFF
--- a/lib/contentful.js
+++ b/lib/contentful.js
@@ -87,27 +87,8 @@ export function createClient (params) {
   // Append environment to baseURL
   http.defaults.baseURL = getGlobalOptions().environmentBaseUrl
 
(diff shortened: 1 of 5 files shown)


marcolink:
I think it would be good to manually add a little "how to upgrade" guide to the readme for the upcoming version, wdyt?

marcolink (approved):

PR #909, opened 2021-09-21
Fully type query helpers

marcolink:
## Todos

-   [x] Implemented feature


lib/contentful.ts
```diff
@@ -7,7 +7,7 @@
 
 import axios from 'axios'
 import { createHttpClient, getUserAgentHeader } from 'contentful-sdk-core'
-import { HttpClientInstance } from './types'
+import { AxiosInstance } from 'contentful-sdk-core/dist/types/types'
```
damienxy, line 10: why not `import { AxiosInstance } from 'contentful-sdk-core'`?

damienxy (approved):

PR #920, opened 2021-10-01
fix(types): add optional environment SpaceLink to Sys interface

mtchalim:
Raw CDA responses include an `environment` field in each entity's `sys` field, which is not currently reflected in the `Sys` typedef.  This PR introduces `environment` as an optional `SpaceLink` in the `Sys` interface, which should better support typing CDA response entities.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -159,6 +159,9 @@ export interface Sys {
     space?: {
         sys: SpaceLink;
     };
+    environment?: {
+        sys: SpaceLink;
+    };
     contentType: {
         sys: ContentTypeLink;
     };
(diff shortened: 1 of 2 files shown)


ruderngespra:
Thanks for reporting this! As this is an EnvironmentLink with a different `linkType` I [slightly adjusted](https://github.com/mtchalim/contentful.js/pull/1) your suggestion (and restructured the other EntityLink types). Let me know if this makes sense to you!

mtchalim:
> Thanks for reporting this! As this is an EnvironmentLink with a different `linkType` I [slightly adjusted](https://github.com/mtchalim/contentful.js/pull/1) your suggestion (and restructured the other EntityLink types). Let me know if this makes sense to you!

Absolutely, thank you!

marcolink (approved):
That looks great - 🚢

PR #921, opened 2021-10-01
Merge master into experimental

damienxy:
## Summary

Bring experimental branch back in sync with current state of master. 
WIP because there are still some TODOs left


lib/types/entry.ts
```diff
@@ -62,6 +65,7 @@ export interface LocalizedEntry<Fields extends FieldsType, Locale extends Locale
       [LocaleName in Locale]?: Fields[FieldName]
     }
   }
+  metadata: Metadata
```
marcolink, line 68: Something we have to test. I think `metadata` is not available for `Link`

lib/types/tag.ts
```diff
@@ -0,0 +1,23 @@
+import { ContentfulCollection } from './collection'
+import { EntryFields } from './entry'
+import { EnvironmentLink, SpaceLink, UserLink } from './link'
+import { BaseSys } from './sys'
+
+export interface TagSys extends BaseSys {
```
marcolink, line 6: we could also extend `EntitySys` and omit the `locale` field

lib/types/entry.ts
damienxy: It just looks like this is in `Link` due to GH diffing view. But it's actually on `LocalizedEntry` :-)

PR #983, opened 2021-11-09
docs(api-methods): fix example of getEntries

andreascful:
Fix the example code for getEntries to prevent syntax error

## Summary

Previous example code for the getEntries API method led to syntax error when copied 1:1. 
Fixed the example code.


marcolink (approved):
🚢

PR #1001, opened 2021-12-01
fix: contentful-sdk-core v6.10.4

johanneswuerbach:
Ensure https://github.com/contentful/contentful-sdk-core/pull/239 is being used.


marcolink (approved):
:shipit:

PR #1043, opened 2022-01-17
220117/fix user agent header issue

ruderngespra:
### DONE
- Header generator picks up the `__VERSION__` var, both type tests as well as header tests pass.
- Manually test that version numbers are actually picked up.
- Investigate how to avoid the `global.d.ts` duplication.
--> In order to avoid uplication, we use ts triple slash reference.


marcolink (approved):
excellent 🎉

PR #1047, opened 2022-01-19
Ts/fix builds

andreascful:
Implements parcel as bundler. Types are being created with tsc.


package.json
```diff
@@ -3,10 +3,17 @@
   "description": "Client for Contentful's Content Delivery API",
   "version": "0.0.0-determined-by-semantic-release",
   "homepage": "https://www.contentful.com/developers/documentation/content-delivery-api/",
-  "main": "./dist/contentful.node.js",
-  "browser": "./dist/contentful.browser.js",
-  "modules": "./dist/contentful.browser.js",
+  "source": "./lib/index.ts",
+  "main": "./dist/contentful.cjs.js",
+  "module": "./dist/contentful.esm.mjs",
```
marcolink, line 8: Why is this defined as `*.mjs` - should be `*.js`right?

package.json
```diff
@@ -3,10 +3,17 @@
...
+  "exports": {
+    "import": "./dist/contentful.esm.mjs", 
+    "require": "./dist/contentful.cjs.js"
+  },
   "types": "./dist/types/index.d.ts",
+  "targets": {
```
marcolink, line 14: Is this the relevant part for Parcel to NOT compile the type definitions?

build.mjs
```diff
@@ -19,20 +19,22 @@ const build = (options) => esbuild
 const browserConfig = {
```
marcolink, line 19: Do we need this `esbuild` build file anymore?

package.json
```diff
@@ -34,11 +41,12 @@
...
   "scripts": {
     "clean": "rimraf dist",
-    "build": "node build.mjs && npm run build:types",
+    "build": "npm run clean && npx parcel build --no-source-maps && tsc --emitDeclarationOnly",
+    "build:dev": "npm run clean && npx parcel build && tsc --emitDeclarationOnly",
     "build:types": "tsc --declaration --emitDeclarationOnly",
     "postbuild": "npm run check:browser && npm run check:node && npm run test:size",
-    "check:browser": "es-check es2017 --module ./dist/contentful.browser.js",
-    "check:node": "es-check es2017 ./dist/contentful.node.js",
+    "check:browser": "es-check es2017 --module ./dist/contentful.esm_.js",
```
marcolink, line 48: where would we generate this output file?

package.json
```diff
@@ -68,6 +76,8 @@
   "devDependencies": {
     "@contentful/browserslist-config": "^2.0.0",
     "@contentful/rich-text-types": "^15.3.6",
+    "@parcel/packager-ts": "^2.0.0-nightly.982",
```
marcolink, line 79: It would be great to have a comment in the PR description why we pinning a nightly build here ;)

package.json
ruderngespra: @marcolink Why would you want to take away the more specific `*.mjs` here, isn't this just a helping guideline for anyone or any tool using this module (which is anyways an ES module)?

Taking this thought from the parcel docs (https://parceljs.org/getting-started/library/#commonjs-and-es-modules): 
image

build.mjs
ruderngespra: I don't think we do and will remove it.

package.json
ruderngespra: But not sure, we should probably be aligned in the structure of all bundled files here 🤔

ruderngespra: And also just saw our internal alignment doc ...

ruderngespra: Okay, I saw that we aligned on always using `*.js`, so I'll change that, but just wonder what it the benefit compared to using the `cjs` and `mjs` endings?

package.json
ruderngespra: As the nightly build didn't solve our problem, I think we should use `@latest` instead.

package.json
ruderngespra: Why are we not using   `./dist/contentful.esm.js` here and +1 to Marcos question.

package.json
```diff
@@ -34,11 +41,12 @@
   "license": "MIT",
   "scripts": {
     "clean": "rimraf dist",
-    "build": "node build.mjs && npm run build:types",
+    "build": "npm run clean && npx parcel build --no-source-maps && tsc --emitDeclarationOnly",
```
ruderngespra, line 44: Why are we using `npx` here, I think this should take up the local version of parcel even without it.

package.json
marcolink: @ruderngespra I actually think it's a typo - we don't create such a file.

ruderngespra: 👍🏻 Removed it.

package.json
ruderngespra: It seems like both `targets` as well as `types` isn't relevant anymore as we don't create types with the help of parcel but rely exclusively on tsc.

marcolink: but `types` is pointing the consuming package to the types definitions, and therefore it's an essential field :)

ruderngespra: Ah, thanks, will revert then.

ruderngespra: And yes, seems that in that case we also need the `targets` property again :)

PR #1059, opened 2022-01-25
Chore/add webpack

ruderngespra:
### done
- creates four bundles based on `webpack.config.js`, two of them minified

### todos
- test in local projects
- targets should ideally fully rely on `browserslist`, but the `"maintained node versions"` flag is not supported by webpack.
- bundles have significantly increased size, can they be further optimized, double check with old config?
- Does the automated release process still work with the updated scripts?
  - Should work, as it relies on the same `npm run build` process + env variable names haven't been changed. But how to test?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,6 +2,7 @@ module.exports = {
   plugins: ['standard', 'promise'],
   env: {
     jest: true,
+    node: true,
   },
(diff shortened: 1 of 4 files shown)


marcolink (approved):
Looks good, do we know why the bundle size increased?

PR #1075, opened 2022-02-07
Chore/add postbuild demos

andreascful:
Adds postbuild tests for two demo projects (node and browser) to check if the created bundles can actually be consumed by clients.

Thank you for the suggestion @marcolink


test/build/browser/index.test.js
```diff
@@ -0,0 +1,18 @@
...
+let browser, page
+
+jest.setTimeout(10000)
+
+beforeEach(async () => {
+  browser = await puppeteer.launch()
+  page = await browser.newPage()
+  await page.goto(`file:${path.join(__dirname, './public/index.html')}`)
+  await page.waitForTimeout(4000)
+})
```
marcolink, line 13: We could clean up the browser again by calling

```js
await browser.close();
```

marcolink:
I think each test build project should have its own package.json. it should have `contentful.js`  as a normal dependency. 
Before the build step, we can override the dependency via our local output via [relative-deps](https://github.com/mweststrate/relative-deps). We want to let the local build decide what field it picks up from our package.json, and not hard code the compiled output.

andreascful:
> I think each test build project should have its own package.json. it should have `contentful.js` as a normal dependency. Before the build step, we can override the dependency with our local output via [relative-deps](https://github.com/mweststrate/relative-deps). We want to let the local build decide what field it picks up from our package.json, and not hard code the compiled output.

At the moment the direct reference is only happening for the browser variant. I think this is a legitimate behaviour since we also recommend the usage via script tag in the README.

The node project only references the whole folder that contains contentful.js and decides which bundle to pick by looking at the package.json. Or did you mean something else?

test/output-integration/browser/package.json
```diff
@@ -0,0 +1,16 @@
+{
```
marcolink, line 1: 👍

.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - run: npm run test:unit
   integration:
     docker:
-      - image: cimg/node:16.11
+      - image: cimg/node:16.13.2-browsers
```
marcolink, line 26: Theoretically, we could make the build tests their own step, then we only have to use the browser image for that step, right?

andreascful: Just to understand you correctly: That would mean that we'd also need to change the chain of commands in the package.json (since atm the tests are executed automatically bei `npm run build`), right? Otherwise those would also execute with the the "lint" step for example.
I am not sure if I'd be a big fan of that though since imho when I run `npm run build` as a dev I want those tests to be executed automatically to see if the output is okay.

What I will definitely change though is to remove the "-browsers" from the steps that don't include the build command.

marcolink (approved):

PR #1080, opened 2022-02-10
Typescript/update dependencies

marcolink:
bump direct depdencies


lib/query-builder.ts
```diff
@@ -0,0 +1,98 @@
...
+  homeLocation: EntryFields.Location
+}
+
+type MyEntry = {
+  sys: EntrySys
+  fields: MyEntryFields
+}
+
+/*
+const query: Query<MyEntry> = {
```
andreascful, line 78: Should the comment stay or should it better be removed?

lib/query-builder.ts
```diff
@@ -0,0 +1,98 @@
+import { ConditionalPick } from 'type-fest'
```
andreascful, line 1: It looks to me like the file is not used/required anywhere. Does it fit into this PR?

marcolink: nope - good catch! this is my local laboratory - will remove it 👍

andreascful (approved):
🚢 !!

PR #1086, opened 2022-02-14
Restrict locale param from clients

Vinz93:
## Implementation changes  

- **Default** **client** shows a warning if `locale: '*'` param is provided, pointing to the new dedicated client 

- **Localized** **client** throws an error if `locale` property is provided

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -488,6 +488,9 @@ export default function createContentfulApi<OptionType>(
   async function getEntriesWithLinkResolution<Fields>(
     query: EntriesQueries<Fields> = {}
   ): Promise<EntryCollectionWithLinkResolution<Fields>> {
+    if (query.locale === '*') {
+      console.warn('If you want to fetch all the locales, we recommend to use the .withAllLocales')
+    }
     return internalGetEntries<EntryCollectionWithLinkResolution<Fields>>(query, true)
   }
 
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -485,7 +485,7 @@ export default function createContentfulApi<OptionType>(
   ): Promise<EntryWithLinkResolution<Fields>> {
     if (query.locale === '*') {
       console.warn(
-        'If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales'
+        `If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales instead of the locale='*' parameter`
       )
     }
     return internalGetEntry<EntryWithLinkResolution<Fields>>(id, query, true)
@@ -496,7 +496,7 @@ export default function createContentfulApi<OptionType>(
   ): Promise<EntryCollectionWithLinkResolution<Fields>> {
     if (query.locale === '*') {
       console.warn(
-        'If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales'
+        `If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales instead of the locale='*' parameter`
       )
     }
     return internalGetEntries<EntryCollectionWithLinkResolution<Fields>>(query, true)
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -484,7 +484,7 @@ export default function createContentfulApi<OptionType>(
   ): Promise<EntryWithLinkResolution<Fields>> {
     if (query.locale === '*') {
       console.warn(
-        'If you want to fetch all the locales, we recommend you to use client.withAllLocales'
+        'If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales'
       )
     }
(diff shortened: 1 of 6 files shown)


lib/create-contentful-api.ts
```diff
@@ -36,7 +36,7 @@ import { EntryQueries } from './types/query/query'
 import { FieldsType } from './types/query/util'
 import normalizeSelect from './utils/normalize-select'
 import resolveCircular from './utils/resolve-circular'
-import validateTimestamp from './utils/validate-timestamp'
+import validateTimestamp, { ValidationError } from './utils/validate-timestamp'
```
damienxy, line 39: I suggest to move `ValidationError` out of validate-timestamp.ts into its own file, as it's now used for more than that

marcolink, line 39: We should move `ValidationError` to its own file.

lib/create-contentful-api.ts
```diff
@@ -488,6 +488,11 @@ export default function createContentfulApi<OptionType>(
   async function getEntriesWithLinkResolution<Fields>(
     query: EntriesQueries<Fields> = {}
   ): Promise<EntryCollectionWithLinkResolution<Fields>> {
+    if (query.locale === '*') {
+      console.warn(
+        'If you want to fetch all the locales, we recommend you to use client.withAllLocales'
```
damienxy, line 493:
```suggestion
        'If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales'
```

marcolink (commented):
Should we have the same behavior for `getEntry`?

lib/create-contentful-api.ts
```diff
@@ -511,6 +516,9 @@ export default function createContentfulApi<OptionType>(
   >(
     query: EntriesQueries<Fields> = {}
   ): Promise<EntryCollectionWithAllLocalesAndWithLinkResolution<Fields, Locales>> {
+    if (query.locale) {
+      throw new ValidationError('locale', 'locale parameter is not allowed')
```
damienxy, line 520:
```suggestion
      throw new ValidationError('locale', 'The `locale` parameter is not allowed')
```

lib/create-contentful-api.ts
```diff
@@ -550,6 +558,9 @@ export default function createContentfulApi<OptionType>(
   >(
     query: EntriesQueries<Fields> = {}
   ): Promise<EntryCollectionWithAllLocalesAndWithoutLinkResolution<Fields, Locales>> {
+    if (query.locale) {
+      throw new ValidationError('locale', 'locale parameter is not allowed')
```
damienxy, line 562:
```suggestion
      throw new ValidationError('locale', 'The `locale` parameter is not allowed')
```

damienxy (changes requested):
Nice! 
We still need to add the same warnings/errors to the `getEntry` method

lib/create-contentful-api.ts
```diff
@@ -482,12 +483,22 @@ export default function createContentfulApi<OptionType>(
...
+    }
     return internalGetEntry<EntryWithLinkResolution<Fields>>(id, query, true)
   }
 
   async function getEntriesWithLinkResolution<Fields>(
     query: EntriesQueries<Fields> = {}
   ): Promise<EntryCollectionWithLinkResolution<Fields>> {
+    if (query.locale === '*') {
+      console.warn(
+        'If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales'
```
damienxy, line 499: Same comment as above

```suggestion
        'If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales instead of the locale='*' parameter
```

lib/create-contentful-api.ts
```diff
@@ -482,12 +483,22 @@ export default function createContentfulApi<OptionType>(
     id: string,
     query: EntryQueries = {}
   ): Promise<EntryWithLinkResolution<Fields>> {
+    if (query.locale === '*') {
+      console.warn(
+        'If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales'
```
damienxy, line 488: [sorry for the long text, it's just there to explain my suggested change ✏️ ]

Just thought about this warning and the error below. I think at the moment we risk the flow being:

1/ user uses `client.getEntries({locale: '*'})`
2/ warning: we recommend you to use `client.withAllLocales`
3/ user changes to `client.withAllLocales.getEntries({locale: '*'})`
4/ error: locale not allowed
5/ user is confused about mixed messages

To mitigate I think we can make the warning more specific so when people see it in the console they can go back to the code and know what to change:

```suggestion
        'If you want to fetch an entry in all existing locales, we recommend you to use client.withAllLocales instead of the locale='*' parameter'
```

damienxy (approved):
Another small text change suggestion, otherwise 💯

lib/create-contentful-api.ts
Vinz93: I agree, great example to show me the confusion :)

marcolink (approved):

ruderngespra (approved):

PR #1088, opened 2022-02-15
Restrict usage of unresolved param

ruderngespra:
- add tests for resolveLinks warnings and errors ✅ 
- implement according to test structure ✅ 
- Double check potential "false positive" warnings when parameter is set not by the user but programmatically. ✅ 

Should be merged into `next-typescript` after [feat/localized-client-restrict-locale-param](https://github.com/contentful/contentful.js/tree/feat/localized-client-restrict-locale-param).

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -62,10 +62,10 @@ export interface ClientWithLinkResolution extends BaseClient {
 export interface ClientWithoutLinkResolution extends BaseClient {
   getEntry<Fields extends FieldsType>(
     id: string,
-    query?: EntryQueries
+    query?: EntryQueries & { resolveLinks?: never }
   ): Promise<EntryWithoutLinkResolution<Fields>>
   getEntries<Fields extends FieldsType>(
-    query?: EntriesQueries<Fields>
+    query?: EntriesQueries<Fields> & { resolveLinks?: never }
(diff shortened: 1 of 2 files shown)


lib/create-contentful-api.ts
```diff
@@ -488,6 +488,11 @@ export default function createContentfulApi<OptionType>(
         `If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales instead of the locale='*' parameter`
       )
     }
+    if ('resolveLinks' in query) {
+      console.warn(
+        'If you do not want to resolve links, we recommend you to use client.withoutLinkResolution'
```
damienxy, line 493:
```suggestion
        'If you do not want to resolve links, we recommend you to use client.withoutLinkResolution instead of the resolveLinks=false parameter'
```

lib/create-contentful-api.ts
```diff
@@ -499,6 +504,11 @@ export default function createContentfulApi<OptionType>(
...
```
damienxy, line 509:
```suggestion
        'If you do not want to resolve links, we recommend you to use client.withoutLinkResolution instead of the resolveLinks=false parameter'
```

damienxy (approved):

Vinz93 (approved):

PR #1097, opened 2022-02-17
chore: [] registed in internal backstage

eamodeorubio:
## Summary

Registed this library in our internal software catalog

## Description

Add backstage meta descriptor file

## Motivation and Context

Internal maintenance.


marcolink (approved):
excellent! - will add the remaining OSS libraries in a next PR

PR #1098, opened 2022-02-17
feat(client): expose chain options

marcolink:
This can be helpful for situations sub clients are passed around and you want to check the current chain.


marcolink:
Not relevant right now

PR #1081, opened 2022-02-10
Implement client modifier chaining

damienxy:
## Summary

Turning client chaining prototype into real implementation. 

## Description

Ready for review. Implementation is not done yet (see Todos), but we'll merge into `next-typescript` now and do the rest based off of that. This unblocks other tasks and prevents several WIP branches being branched off of each other. 

Done so far:
- Added an additional `makeClient` layer when creating contentful API. We can still change the structure/flow.
- Fixed tests broken by new structure
- Chained clients return corresponding response types for `getEntries` and `getEntry`. 
- Added tests for all client chains for `getEntries` and `getEntry` implementation.

## Todos

List of things to do (non-exhaustive):
- add third chain (`withoutUnresolvedLinks`)
- add tests for third chain etc
- write dedicated JS doc for each `getEntries` & `getEntry` implementation
- decide how to deal with config and defaults (= proper overrides)
- deal with TODO comments in the code

PR #1102, opened 2022-02-21
Implement client modifier chaining (Part 2 - third chain)

damienxy:
## Summary

Continuation of PR #1081: Turning client chaining prototype into real implementation. 
This PR includes the third chain (`withoutUnresolvableLinks`).

## Description

Done in this PR:
- Added third chain (`withoutUnresolvedLinks`)
- Added tests for third chain (in `tests.test.ts` file, should be extended and moved to `getEntries` and `getEntry` test files)
- Made type names more explicit due to adding of third chain
- Implemented `EntryWithLinkResolutionAndWithoutUnresolvableLinks` and `EntryWithAllLocalesAndWithLinkResolutionAndWithoutUnresolvableLinks` types 
- Verified and fixed all Entry type return shapes

## Todos

List of things to do (non-exhaustive):
- ~implement `EntryWithLinkResolutionAndWithoutUnresolvableLinks` and `EntryWithAllLocalesAndWithLinkResolutionAndWithoutUnresolvableLinks` types properly (see TODO comments [in the code](https://github.com/contentful/contentful.js/blob/10a941849ea8cd818ebe64049a848a60829c6b9c/lib/types/entry.ts#L130-L144))~
- ~verify all Entry type return shapes, see "TYPE SHOULD BE" comments [in the code](https://github.com/contentful/contentful.js/blob/feat/implement-client-modifier-chaining-part2/lib/types/entry.ts#L66-L140)~
- ~make config overrides less fragile (see TODO comment in the code)~
- ~add tests for config overrides (there are some already, but not everything is covered)~ [not needed in this PR]
- ~write dedicated JS doc for each `getEntries` & `getEntry` implementation~ [not needed in this PR]
- ~type names are getting really long. Is this too unreadable? Should we change sth?~ [we leave it like this for now]
- ~deal with all the other chained client-related TODO comments in the code.~
- ~integration tests for third chain are in `tests.test.ts` file, should be extended and moved to `getEntries` and `getEntry` test files~ [will be done in seperate PR]


lib/create-contentful-api.ts
```diff
@@ -656,30 +770,44 @@ export default function createContentfulApi<OptionType>(
...
       const entries = await get({
         context: 'environment',
         path: 'entries',
         config: createRequestConfig({ query: normalizeSelect(query) }),
       })
-      return resolveCircular(entries, { resolveLinks, removeUnresolved }) as RType
+
+      return resolveCircular(entries, {
+        resolveLinks: !withoutLinkResolution,
+        removeUnresolved: !!withoutUnresolvableLinks,
```
damienxy, line 809: I would be more specific with the default value, then we also don't need the double negation: 

```suggestion
        resolveLinks: !withoutLinkResolution ?? true,
        removeUnresolved: withoutUnresolvableLinks ?? false,
```

Vinz93: I believe that `resolveLinks` will always have a value, and the `!!` it's just casting to boolean which cover the `undefined` case, I feel that adding default value it's a bit redundant 🤔

damienxy: True, but I think if one of them can be `undefined`, it makes sense to have a default value, and to be consistent I would also give a default value to the other one (I agree it's not really necessary for that one). 
Plus, it implicitly documents the default behavior of the contentful.js library (i.e. returning entries with link resolution and with unresolvable links)

lib/make-client.ts
```diff
@@ -1,26 +1,32 @@
...
+  isClientWithLinkResolutionAndWithUnresolvableLinks,
   isClientWithoutLinkResolution,
+  isClientWithLinkResolutionAndWithoutUnresolvableLinks,
 } from './utils/client-helpers'
 
-function create<OptionsType, ClientType>(
+function create<OptionsType extends ChainOptions, ClientType>(
   { http, getGlobalOptions }: CreateContentfulApiParams,
   options: OptionsType,
   makeInnerClient: (options: OptionsType) => ClientType
```
ruderngespra, line 32: Don't we need an additional

```typescript
  Object.defineProperty(response, 'withoutUnresolvableLinks', {
    get: () => makeInnerClient({ ...options, ... }),
  })
```
in order to make chains like `client.withAllLocales.withoutUnresolvableLinks.getEntry` work?

ruderngespra (commented):
Focused on the "third chain" part, looks great to me, some comments where I also just might not be up to date with latest design decisions, in any case I'd want none of those issues to be understood as blockers for merging to `next-typescript`.

lib/create-contentful-api.ts
```diff
@@ -40,22 +44,24 @@ import validateTimestamp from './utils/validate-timestamp'
...
-  isClientWithLinkResolution,
+  isClientWithLinkResolutionAndWithoutUnresolvableLinks,
+  isClientWithLinkResolutionAndWithUnresolvableLinks,
   isClientWithoutLinkResolution,
 } from './utils/client-helpers'
 
 const ASSET_KEY_MAX_LIFETIME = 48 * 60 * 60
 
-export interface ClientWithLinkResolution extends BaseClient {
+export interface ClientWithLinkResolutionAndWithUnresolvableLinks extends BaseClient {
```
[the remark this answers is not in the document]
damienxy: You're right, all methods will be available on all chains. 
That's intended though, so you can define which kind of client you want once at the beginning (if that's what you want to do) and then use that client for all your calls.

E.g. if you know that you always want entries in all locales, you can set up your project like this:

```js
const baseClient = contentful.createClient()
const client = baseClient.withAllLocales

const contentTypes = await client.getContentTypes() 
const localizedEntries = await client.getEntries()
```

test/integration/tests.test.ts
```diff
@@ -164,14 +165,61 @@ test('Gets entry with without link resolution but with includes', async () => {
...
+})
+test('Gets entry with link resolution and includes, removing unresolvable links via client chain', async () => {
+  const response = await client.withoutUnresolvableLinks.getEntry('4SEhTg8sYJ1H3wDAinzhTp', {
+    include: 2,
+  })
+  expect(response.fields).toBeDefined()
+  expect(response.fields.bestFriend).toBeUndefined()
+})
+
+test('Gets entries with link resolution and includes, removing unresolvable links, overriding client config with client chain', async () => {
```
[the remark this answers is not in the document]
damienxy: exactly 👍

lib/create-contentful-api.ts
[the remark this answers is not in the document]
ruderngespra: Makes sense, thanks :)

lib/make-client.ts
damienxy: you're right! fixed

lib/create-contentful-api.ts
```diff
@@ -509,39 +547,99 @@ export default function createContentfulApi<OptionType>(
         'The use of the `resolveLinks` parameter is discouraged. By default, links are resolved. If you do not want to resolve links, we recommend you to use client.withoutLinkResolution.'
       )
     }
-    return internalGetEntries<EntryCollectionWithLinkResolution<Fields>>(query, true)
+    return internalGetEntries<EntryCollectionWithLinkResolutionAndWithUnresolvableLinks<Fields>>(
+      query,
+      { withoutLinkResolution: false }
```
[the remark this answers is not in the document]
damienxy: Very good point. Let's discuss the different options (in another place, will merge this PR in the meantime).

ruderngespra (approved):
🚢 🎉

Vinz93 (approved):

marcolink (approved):

PR #1134, opened 2022-02-28
Test/user agent header test [SDK-1952]

andreascful:
Adds a postbuild test to check for the existence and correctness of the X-Contentful-User-Agent header.


ruderngespra (approved):
LGTM :) Just a pity that the test is so hard to test. 🤯 

Some generic questions for my understanding:

If I understand the logic correctly, `prepublishOnly` which has a special meaning in the npm world, is only called before the actual publishing? And so we are assuming that whenever a semantic-release is triggered, we check for an actual version number, if not, we are fine with our test default?

How do we know that the `package.json` at the moment that `prepublishOnly` is called has actually already the correct version number to compare with and not the `0-0-0-determined-by-semantic-release`?

Will we add the same for the browser bundle?

andreascful:
@ruderngespra 

> If I understand the logic correctly, prepublishOnly which has a special meaning in the npm world, is only called before the actual publishing? And so we are assuming that whenever a semantic-release is triggered, we check for an actual version number, if not, we are fine with our test default?

Yes, this is correct. You can find more about prepublishOnly here: https://docs.npmjs.com/cli/v8/using-npm/scripts

> How do we know that the package.json at the moment that prepublishOnly is called has actually already the correct version number to compare with and not the 0-0-0-determined-by-semantic-release?

This should always be the case according to the semantic-release documentation: https://semantic-release.gitbook.io/semantic-release/support/faq#how-can-i-use-a-npm-build-script-that-requires-the-package.jsons-version

> Will we add the same for the browser bundle?

Good suggestion. I added those tests.

PR #1138, opened 2022-03-02
feat(params): Remove asterisk as option for locale in client query param [SDK-1979]

andreascful:
Remove * as an option for the locale query param for getEntries and getEntry. The query param `locale='*'` has been deprecated in favor of the client chain `withAllLocales`.

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -519,11 +519,7 @@ export default function createContentfulApi<OptionType>(
     id: string,
     query: EntryQueries = {}
   ): Promise<EntryWithLinkResolutionAndWithUnresolvableLinks<Fields>> {
-    if (query.locale === '*') {
-      console.warn(
-        `If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales instead of the locale='*' parameter.`
-      )
-    }
+    validateLocaleParam(query)
     if ('resolveLinks' in query) {
       console.warn(
         'The use of the `resolveLinks` parameter is discouraged. By default, links are resolved. If you do not want to resolve links, we recommend you to use client.withoutLinkResolution.'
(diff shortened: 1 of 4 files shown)


lib/create-contentful-api.ts
```diff
@@ -520,8 +520,8 @@ export default function createContentfulApi<OptionType>(
     query: EntryQueries = {}
   ): Promise<EntryWithLinkResolutionAndWithUnresolvableLinks<Fields>> {
     if (query.locale === '*') {
-      console.warn(
-        `If you want to fetch entries in all existing locales, we recommend you to use client.withAllLocales instead of the locale='*' parameter.`
+      console.error(
+        `To fetch an entry in all existing locales use client.withAllLocales instead of the locale='*' parameter.`
       )
     }
```
damienxy, line 526: We should throw a ValidationError here (and at the other places):
```js
if (query.locale === '*') {
    throw new ValidationError('locale', 'To fetch an entry in all existing locales use client.withAllLocales instead of the `locale='*'` parameter.')
}
```

damienxy (changes requested):
I think we need this error at more places. 

General rule:

- For `getEntr*` functions including `withAllLocales` in their name: 

```
if (query.locale) {
    throw new ValidationError('locale', 'The `locale` parameter is not allowed')
}
```

  

- For `getEntr*` functions **not** including `withAllLocales` in their name: 

```
if (query.locale === '*') {
    throw new ValidationError('locale', 'To fetch entries in all existing locales, use client.withAllLocales instead of the `locale='*'` parameter.')
}
```

So I think places where it's still missing are (maybe best to double-check): 

  - `getEntryWithLinkResolutionAndWithoutUnresolvableLinks`
  - `getEntriesWithLinkResolutionAndWithoutUnresolvableLinks`
  - `getEntryWithoutLinkResolution`
  - `getEntriesWithoutLinkResolution`

test/integration/tests.test.ts
```diff
@@ -510,7 +510,7 @@ describe('Sync API', () => {
 })
 
 test("Gets entries with linked includes with all locales using locale:'*' parameter", async () => {
```
damienxy, line 512:
```suggestion
test("Gets entries with linked includes with all locales using the withAllLocales client chain", async () => {
```

damienxy (approved):

PR #1140, opened 2022-03-03
Feat: deprecate resolveLinks query param in getEntries/getEntry

damienxy:
## Summary

As link resolution is now determined via client chain modifiers, we no longer support the `resolveLinks` query parameter in the `getEntries` and `getEntry` methods.


lib/create-contentful-api.ts
```diff
@@ -51,13 +50,14 @@ import {
...
   isClientWithoutLinkResolution,
 } from './utils/client-helpers'
+import { validateLocaleParam, validateResolveLinksParam } from './utils/validate-params'
 
 const ASSET_KEY_MAX_LIFETIME = 48 * 60 * 60
 
 export interface ClientWithLinkResolutionAndWithUnresolvableLinks extends BaseClient {
   getEntry<Fields extends FieldsType>(
     id: string,
-    query?: EntriesQueries<Fields>
```
andreascful, line 60: Why don't we need <Fields> here anymore?

lib/create-contentful-api.ts
```diff
@@ -710,6 +658,11 @@ export default function createContentfulApi<OptionType>(
       withAllLocales: false,
     }
   ) {
+    const { withAllLocales } = options
+
+    validateLocaleParam(query, !!withAllLocales)
```
andreascful, line 663: `withAllLocales` instead of `!!withAllLocales` maybe since it's gonna be boolean or undefined anyway

lib/create-contentful-api.ts
damienxy: The query had the wrong type (the one for `getEntries`, but this is `getEntry`)

lib/create-contentful-api.ts
damienxy: I added the `!!` so I we can have `validateLocaleParam` take a boolean as second param (see [here](https://github.com/contentful/contentful.js/pull/1140/files#diff-32b5706a8b1c87b87bdd11d85069a594c56b3aa6459daa7b91aef7a4200d84e1R19))

PR #1151, opened 2022-03-10
feat(types): suggest optional keys

marcolink:
Adds optional (field) keys to known field types.
If a field is optional, we also add generate a mapped type for it.


andreascful (approved):
Checked code and tested on our demo project. Lgtm! 🚢

marcolink:
> Checked code and tested on our demo project. Lgtm! 🚢 Edit: The only thing is that fields typed as "never" will now also show up in the suggestion. However, this doesn't seem to be a big problem in my opinion.

@andreascful Is this in the case of the chains where we explicitly deny certain fields?

PR #1152, opened 2022-03-10
feat(content-types): add query params

marcolink:
Adds the ability to use full-text search for content types collection endpoint.


andreascful (approved):
lgtm! 🚢

PR #1153, opened 2022-03-10
fix(types): Fix available sys query params[SDK-1993]

andreascful:
Fix sys query params for getEntries and getAssets so that only actually functioning params will be suggested.


damienxy (approved):
🚢

PR #1162, opened 2022-03-16
fix: remove --dry-run from semantic-release

damienxy:
## Summary

Removing the --dry-run flag from semantic-release


marcolink (approved):
🚀

PR #1216, opened 2022-04-19
fix(dep): Update fast-copy to 2.1.3

andreascful:
Update fast-copy dependency to version 2.1.3 to fix sourcemap issues caused by it. 

See: https://github.com/planttheidea/fast-copy/pull/65


marcolink (approved):
merci! 🚢

damienxy (approved):

Issue #1272, opened 2022-05-17
v10 beta: TypeScript fails because "export type" is required when re-exporting types with isolatedModules flag


grubersjoe:
## Expected Behavior

I'm using the beta of v10 in a Next.js based page to fetch data roughly like this:

```ts
export interface Fields {
  body: EntryFields.Text;
  // ...
}

const fetch = async (query: EntriesQueries<Fields>) =>
  client
    .withoutLinkResolution
    .getEntries<Fields>({ content_type: 'type' });
```

When running `yarn tsc` I would not expect type errors.

## Actual Behavior

However, I see the following TypeScript error:

```shell
Failed to compile.

./node_modules/contentful/lib/types/query/index.ts:1:10
Type error: Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.

> 1 | export { AssetQueries, EntriesQueries } from './query'
    |          ^
```

This is using the default Next.js `tsconfig.json` that has `isolatedModules` enabled indeed:

```json
{
  "compilerOptions": {
    "target": "es5",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "incremental": true
  },
  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
  "exclude": ["node_modules"]
}
```

## Possible Solution

Using type exports as suggested in `lib/types/query/index.ts` fixes the issue:

```ts
export { type AssetQueries, type EntriesQueries } from './query'
```

This is supported since [TypeScript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html).

## Environment

- **Package Version**: 10.0.0-beta-v10.6
- **Framework**: Next.js v12.1.6
- **TypeScript**: v4.6.4

I could open a merge request if desired.

marcolink:
Hi @grubersjoe 👋 

thanks for reporting this 🚀  

a PR would be great. Feel free to tag me for a quick review!

grubersjoe:
Here you go: https://github.com/contentful/contentful.js/pull/1273 👍

marcolink:
closing this as a fix has been released!

PR #1273, opened 2022-05-18
Use type only exports to support isolatedModules

grubersjoe:
## Summary

See https://github.com/contentful/contentful.js/issues/1272 for context.

@marcolink


marcolink (approved):
@grubersjoe thanks for the contribution 🎉

Issue #1274, opened 2022-05-19
Location datatype wrong


phawxby:
https://github.com/contentful/contentful.js/blob/405cc36a980a08ea7ce9f23f33b6f94488f4c449/index.d.ts#L237-L238

Based on the published datamodel here.
https://www.contentful.com/developers/docs/concepts/data-model/

The lat/lon values are numeric.
<img width="317" alt="image" src="https://user-images.githubusercontent.com/1090602/169305694-7e839070-20a5-4625-813d-5b40c69ef4cf.png">

Vinz93:
Hola @phawxby! You are right, I'll adjust it. Thanks for caching it 🙂

PR #1280, opened 2022-05-20
fix: Location type

Vinz93:
Type adjustment to Location in Entry Field

Fix: [#1274](https://github.com/contentful/contentful.js/issues/1274#issuecomment-1132527411)


marcolink (approved):
🚀

PR #1281, opened 2022-05-20
fix: location type (string -> number)

Vinz93:
Fix location type from `string` to `number`

Related to: [#1274](https://github.com/contentful/contentful.js/issues/1274)


marcolink (approved):
good catch 🎉

PR #1391, opened 2022-08-03
Fix sync type to accept paginate as an option

t-botz:
## Summary

Sync methods accept  options as a second argument with pagination.

## Description

With typescript, compilation fail when doing
```typescript
client.sync(query, { paginate: false })
  ```


marcolink:
Hey @t-botz 👋 ,
We expected this to be solved with the latest [major](https://github.com/contentful/contentful.js/compare/v9.3.5...v10.0.0)  release (which was a complete re-write).
Please feel free to re-open again if the error still occurs.

PR #1411, opened 2022-08-28
Type json more accurately

coryasilva:
## Summary

This PR more accurately types Json fields on Entries.

## Motivation and Context

`[]` is valid json but does not match the type (possible in contentful app).  Furthermore, the current type uses `Record<string, any>` which is too general.


Cyberxon (approved):

marcolink:
With type-fest, we also have a JSON definition in place
https://github.com/sindresorhus/type-fest#json

veu:
We’ve fixed the types and we’re now using the JSON types from type-fest. Thanks for the suggestion, @coryasilva!

PR #1428, opened 2022-09-12
chore(secrets): add vault-secrets

andreascful:
## Summary

Add a vault-secrets file to enable using a secrets management solution


marcolink:
Is there a reason why we didn't add CircleCI policies?

PR #1452, opened 2022-09-19
xspace support beta v10

alicankargin:
## Summary

Write tests for xspace support and update jsdocs:
- Unit tests for ResourceLink
- Integrations tests for ResourceLink
- Add types

You can find the content type for integration tests [here](https://app.contentful.com/spaces/ezs1swce23xe/content_types/catalog/fields).
You can find the entry for  integration tests [here](https://app.contentful.com/spaces/ezs1swce23xe/entries/6yfSzwXo99q8BKzkE5AIKo).

### DIFF
--- a/lib/types/content-type.ts
+++ b/lib/types/content-type.ts
@@ -22,6 +22,13 @@ export interface ContentTypeField {
   type: ContentTypeFieldType
   validations: ContentTypeFieldValidation[]
   items?: FieldItem
+  allowedResources?: ContentTypeAllowedResources[]
+}
+
+interface ContentTypeAllowedResources {
+  type: string
+  source: string
+  contentTypes: string[]
 }
 
 export type ContentTypeFieldType =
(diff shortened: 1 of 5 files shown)


lib/types/content-type.ts
```diff
@@ -22,6 +22,13 @@ export interface ContentTypeField {
   type: ContentTypeFieldType
```
luizfonseca, line 22: does the `type` needs to be update here as well? (e.g. type is `ResourceLink`)

test/unit/mocks.ts
```diff
@@ -91,6 +91,32 @@ const contentTypeMock: ContentType = {
       omitted: false,
       validations: [],
     },
+    {
```
luizfonseca, line 94: same here, might be useful to have the field with `type: 'ResourceLink'`

Bikappa (approved):

PR #1469, opened 2022-09-30
auto generate .nvmrc w/ npm run nvmrc package script

oswish:
## Summary

Auto generate the `.nvmrc` file.

## Description

Auto generate `.nvmrc` file from an npm script that looks at the `engines.node` field in the package file. Remove the current `.nvmrc` file that is in source control, now that it is getting auto-genrated. Update the `engines.node` field to be ">=16", the node version specified in the checked in [file](https://github.com/contentful/contentful.js/blob/master/.nvmrc). Add to `.gitignore` the auto generated file. Include the updated package lock.

## Motivation and Context

A contributor can run `npm run nvmrc; nvm install` to automate the process for getting on the correct version of NodeJS.

## Todos

-   [ ] Update the README file to explain usage & dependency on `nvm`.

*no-visuals


marcolink:
Since this might collide with other parts of our CI pipeline, I will close it for now

PR #1494, opened 2022-10-24
chore/update deps on beta

marcolink:
Dependabot updates did not go through lately so we decided to do it with a single batch update!

This update also ensures zero vulnerabilities


crissto (approved):
I guess dependabot doesn't work because it isn't `main` right? Looks great 😁

PR #1569, opened 2022-12-12
Added embedded-asset-{inline|block} types

### LINKED ISSUE #1534: Missing type for wysiwyg content
Type RichTextContent aka wysiwyg entry content is represented like this: 
image

I am trying to write a conditional for assets: 
image

It looks like 'embedded-asset-block' needs to be added to RichTextNodeType in index.d.ts like this (I added it for the pic):
image

This is what needs to be updated: https://github.com/contentful/contentful.js/blob/d25879c40058b237b268e42de9ba9354f6616b70/index.d.ts#L263

DKirwan:
## Summary

Adds the `embedded-asset-inline` and `embedded-asset-block` `RichTextNodeType`s. 

## Description

I added two new types that I had been receiving from the API after embedding an asset in a rich text property.

## Motivation and Context

Fixes #1534 

## Todos

-   [x] Implemented feature

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -211,6 +211,8 @@ export interface FieldValidation {
         'entry-hyperlink'?: FieldValidation[];
         'embedded-entry-block'?: FieldValidation[];
         'embedded-entry-inline'?: FieldValidation[];
+        'embedded-asset-block'?: FieldValidation[];
+        'embedded-asset-inline'?: FieldValidation[];
     };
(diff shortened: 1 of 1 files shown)


z0al:
Thank you for contributing @DKirwan. However, the original issue is fixed by #1617 using a different approach. 

I also need to mention that we don't support `embedded-asset-inline` in the rich text so that won't be there. Please update to the latest version

Issue #1601, opened 2023-01-11
Package does not build to ESM, files are in CJS format


Techn1x:
PR https://github.com/contentful/contentful.js/pull/1602

## Expected Behavior

contentful package should be available as ES Modules

## Actual Behavior

contentful package compiles to CommonJS - even though there is a `dist/es-modules/` directory, the files in there are compiled to CJS rather than ESM format

Notice this file is inside the `dist/es-modules` directory, but the code uses `require()` which is a CJS thing.
![Screenshot from 2023-01-11 16-33-06](https://user-images.githubusercontent.com/1049837/211734684-c25735d7-2e38-42bf-9feb-e9e203783fbb.png)

## Possible Solution

Some kind of webpack configuration change in https://github.com/contentful/contentful.js/blob/master/webpack.config.js
~Sorry that's the best I've got at this stage~
See PR https://github.com/contentful/contentful.js/pull/1602

## Steps to Reproduce

1. Pull the `contentful` package from git, master branch
2. Run `npm run clean && npm run build:modules`
3. Notice that the files in `dist/es-modules` are in CJS format

## Context

I am trying to consume this package as ESM

I would like to point out that `contentful-sdk-core` seems to compile as expected - here's the `index.es-modules.js` file for that package for example
![Screenshot from 2023-01-11 16-35-38](https://user-images.githubusercontent.com/1049837/211735068-11b3b515-735d-4b0b-bcfe-0a10c606b6a5.png)

## Environment

node 16.14.0
ubuntu 22.04

Techn1x:
@marcolink just bringing this one to your attention, not sure if anyone's got notifications for new issues & PRs on this repo

PR #1602, opened 2023-01-11
fix: compile ESM modules

Techn1x:
resolves https://github.com/contentful/contentful.js/issues/1601

---

It seems counterintuitive, but;
```
@babel/preset-env: The 'modules' option must be one of 
 - 'false' to indicate no module processing
 - a specific module type: 'commonjs', 'amd', 'umd', 'systemjs' - 'auto' (default) which will automatically select 'false' if the current
   process is known to support ES module syntax, or "commonjs" otherwise
```

Meaning by setting it to false, it won't try to process modules into CJS etc.

The contentful-sdk-core package does this as well
https://github.com/contentful/contentful-sdk-core/blob/c7841fe403e1a2f62a45b9777857a4180676afdd/rollup.config.js#L21-L26

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -44,6 +44,7 @@ module.exports = function(api) {
       },
       modules: {
         presets: [['@babel/preset-env', {
+          modules: false,
           targets: {
             browsers: mainSupportedBrowsers,
             node: '12'


.babelrc.js
```diff
@@ -44,6 +44,7 @@ module.exports = function(api) {
       },
       modules: {
         presets: [['@babel/preset-env', {
+          modules: false,
```
Techn1x, line 47: It seems counterintuitive, but;
```
@babel/preset-env: The 'modules' option must be one of 
 - 'false' to indicate no module processing
 - a specific module type: 'commonjs', 'amd', 'umd', 'systemjs' - 'auto' (default) which will automatically select 'false' if the current
   process is known to support ES module syntax, or "commonjs" otherwise
```

Meaning by setting it to false, it won't try to process modules into CJS etc.

marcolink:
Thanks for your contribution and fixing this @Techn1x 👍 
I will trigger the release on Monday!

marcolink:
Thanks @Techn1x - I'm about to fix it :)

marcolink:
@Techn1x here you go :) 
https://github.com/contentful/contentful.js/releases/tag/v9.2.17

PR #1627, opened 2023-01-19
chore/add vault secrets

marcolink:
Adds missing vault secrets config

### DIFF
--- a/.contentful/vault-secrets.yaml
+++ b/.contentful/vault-secrets.yaml
@@ -1,0 +1,9 @@
+version: 1
+services:
+  github-action:
+    policies:
+      - dependabot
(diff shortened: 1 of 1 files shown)


.contentful/vault-secrets.yaml
```diff
@@ -0,0 +1,9 @@
+version: 1
+services:
+  github-action:
+    policies:
+      - dependabot
+  circleci:
+    policies:
+      - npm-read
```
andreascful, line 8: Afaik this should not be required separately since the `semantic-release-ecosystem` already has thee required NPM token

marcolink:
solved by #1613

PR #1650, opened 2023-01-26
feat: add allowed resources to Field

alicandko:
## Summary

Adding allowedFields to content type Field interface. This was omitted previously by mistake. It is available in beta-v10: https://github.com/contentful/contentful.js/blob/beta-v10/lib/types/content-type.ts#L25.


marcolink (approved):
Looks good 🚢

PR #1673, opened 2023-02-08
feat: resolve resource links [DANTE-852]

Silhoue:
Add support for client-side ResourceLinks resolution (basic implementation).

`.getEntry()` and `.getEntries()` functions now resolve references to entries from other spaces as they would for regular references, respecting existing `resolveLinks`, `removeUnresolved` and `include` parameters.

Known issues/follow-up tasks are documented in the ticket.


lib/utils/resource-links.js
```diff
@@ -0,0 +1,60 @@
+function isResourceLink (field) {
+  return field && field.sys && field.sys.type === 'ResourceLink'
+}
```
BasKiers, line 3: I would make this check a bit more resilient to prevent bugs when the customer is relying on this SDK version and we release new `ResourceLink` types.
```suggestion
function isContentfulEntryResourceLink (field) {
  return field && field.sys && field.sys.type === 'ResourceLink' && field.sys.linkType === 'Contentful:Entry'
}
```

lib/utils/resource-links.js
```diff
@@ -0,0 +1,60 @@
...
+function getResourceLinkFields (data) {
+  const items = [].concat(...(data.items || []), ...Object.values(data.includes || {}).flat())
+  const fields = items.flatMap(({ fields }) => Object.values(fields).flat())
+  return fields.filter(isResourceLink)
+}
+
+function parseResourceLink (resourceLink) {
+  const urn = resourceLink.sys.urn
+  // For now just assume every resourceLink has type Contentful:Entry
+  const [, spaceId, , entryId] = urn.split('/')
```
BasKiers, line 14: This was written pretty hastily (by me), is it worth the effort to use a better urn parse function?

lib/utils/resource-links.js
```diff
@@ -0,0 +1,60 @@
+function isResourceLink (field) {
+  return field && field.sys && field.sys.type === 'ResourceLink'
+}
+
+function getResourceLinkFields (data) {
+  const items = [].concat(...(data.items || []), ...Object.values(data.includes || {}).flat())
+  const fields = items.flatMap(({ fields }) => Object.values(fields).flat())
+  return fields.filter(isResourceLink)
```
BasKiers, line 8: This code breaks if the `locale=*` parameter is set as it assumes the field response does not include the localisation map

lib/utils/resource-links.js
```diff
@@ -0,0 +1,60 @@
...
+
+async function getEntitiesFromSpace (spaceId, entries, extraApis, include) {
+  const apiInstance = extraApis[spaceId]
+  if (!apiInstance) {
+    return []
+  }
+
+  const entryIds = entries.map(({ entryId }) => entryId)
+
+  let response = { items: [] }
```
BasKiers, line 34: Does not need to be a `let` but can be instantiated as a `const` on line 36
```suggestion
```

lib/utils/resource-links.js
```diff
@@ -0,0 +1,60 @@
...
+  const uniqueEntryIds = [...new Set(entryIds)]
+  response = await apiInstance.getEntries({ 'sys.id[in]': uniqueEntryIds.join(','), include }).catch(() => ({}))
+  return response.items || []
```
BasKiers, line 37: We should probably not only return the `items` but also the `includes` from the response (for assets and same-space Links). This would probably mean changing the return type which will require some refactoring in `addResourceLinksToIncludes`
```suggestion
  return {...(response.includes || {}), Entry: {...((response.includes && response.includes.Entry) || []), ...(response.items || [])}
```

lib/utils/resource-links.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export async function addResourceLinksToIncludes (data, extraApis = {}, include = 1) {
```
BasKiers, line 40: FYI: This function is not written recursively. You'll only be able to resolve 1 layer of x-space ResourceLinks.

lib/utils/resource-links.js
BasKiers: I was incorrect, I forgot that the `apiInstance` will call it's own `wrapEntryCollection` function to copy the includes into the field key

lib/utils/resource-links.js
Silhoue: It is, `addResourceLinksToIncludes` calls `getEntitiesFromSpace` which calls `getEntries` which calls `addResourceLinksToIncludes` again

lib/utils/resource-links.js
Silhoue: Good point, I'll look into this

lib/utils/resource-links.js
BasKiers: You are right. I made the same mistake in my other comment where I forgot that every api instance is doing its own recursive resolving

Silhoue: I added a test case for it

lib/utils/resource-links.js
Silhoue: Since we've already checked that this is a `Contentful:Entry` ResourceLink it should be fine. If we were to change something in the crn pattern it will stop working anyway in the [resolve-response](https://github.com/contentful/contentful-resolve-response/blob/b44ce875c33048d2c19c9a7ca7dd341bd1a237da/index.js#L63).

test/integration/resource-links-test.js
```diff
@@ -0,0 +1,137 @@
...
+ *     -- count
+ */
+
+const client = contentful.createClient({
+  accessToken: '2O_YlI7Spwm16XXw-VZw9vxIIlUOeHLuzWwKXn3KMvo',
+  space: 'z788fz497ijs',
+  additionalTokens: {
+    mbfd18am5xhd: 'tZO0vih_LPefnfNhCDARaRjfiSNXLeJMYRvFe7U-6Yk',
+    p9sqh626r1fo: 'CagovFaZwA7Dta0chjKBU2iU4ASND0TLgCgQyep-Xnk'
+  }
```
luizfonseca, line 34: are those real tokens or mocked ones?

Silhoue: Real CDA tokens to the testing space. This is how integration tests are currently working here.

luizfonseca:
Could you add a brief description to the PR indicating what this is adding and it works?

Maybe some points to clarify:

- Does it keep resolving cross-space links until it finds nothing?
- Can the user add a limit?

lib/utils/resource-links.js
```diff
@@ -0,0 +1,64 @@
...
+async function getEntitiesFromSpace (spaceId, entries, extraApis, include, locale) {
+  const apiInstance = extraApis[spaceId]
+  if (!apiInstance) {
+    return []
+  }
+
+  const entryIds = entries.map(({ entryId }) => entryId)
+  const uniqueEntryIds = [...new Set(entryIds)]
+  const response = await apiInstance.getEntries({ 'sys.id[in]': uniqueEntryIds.join(','), include, locale }, { wrapEntryCollection: false }).catch(() => ({}))
+  return getEntries(response)
```
BasKiers, line 41: I believe this will cause `Asset` fields from the `includes` of the `ResourceLink` object to be discarded

BasKiers: I would suggest putting `items` in the `includes.Entry` array and returning the `includes` object as a whole. In `addResourceLinksToIncludes` you can rewrite the merging logic to merge all `includes` keys e.g.
```javascript
const resourceLinkEntries = (await Promise.all(resourceLinkResponses)).flat() // {[type: String]: Record<string, any>[]}[]
const includesObjects = [...resourceLinkEntries, data.includes || {}]
const includesKeys = new Set(includesObjects.flatMap(({includes = {}}) => Object.keys(includes))
return {
  ...data,
  includes: Array.from(...includesKeys).reduce((includes, key) => ({
    ...includes,
    [key]: includesObjects.flatMap((incl) => incl[key] || [])
  }), {})
}
```

luizfonseca:
You have to update these types as well: https://github.com/contentful/contentful.js/blob/6451a5c286f2243d233008cfb49b5eae1682dd9d/index.d.ts#L16

lib/utils/resource-links.js
```diff
@@ -0,0 +1,64 @@
...
+  const resourceLinkResponses = Object.entries(entitiesBySpace).map(([spaceId, entries]) =>
+    getEntitiesFromSpace(spaceId, entries, extraApis, include - 1, locale)
+  )
+  const resourceLinkEntries = (await Promise.all(resourceLinkResponses)).flat()
+
+  return {
+    ...data,
+    includes: {
+      ...data.includes,
+      Entry: [...((data.includes && data.includes.Entry) || []), ...resourceLinkEntries]
```
luizfonseca, line 61: Q: similar to @BasKiers  comment, I am querying using the client lib but I don't get any of the includes of the other request (e.g. assets etc)

Is it something we need to support now or a follow-up?

Silhoue: I pushed a fix for including the assets. Does that fix the issue you noticed or is it something else?

luizfonseca: Just checked; fixed 👍🏽

ADVANCED.md
```diff
@@ -62,6 +62,29 @@ The link resolution is applied to one level deep by default. If you need it to b
 
 By default, the SDK will keep links, which could not get resolved, in your response. If you want to completely remove fields which could not be resolved, set `removeUnresolved: true` in the configuration options.
 
+### Links to other spaces (beta)
+
+`getEntry` and `getEntries` endponts can now resolve links to entries from other spaces as they do for regular links. The `resolveLinks`, `removeUnresolved` and `include` parameters affect both types of links.
```
Cyberxon, line 67: endpoints*

ADVANCED.md
```diff
@@ -62,6 +62,29 @@ The link resolution is applied to one level deep by default. If you need it to b
...
+
+To use this functionality provide `additionalTokens` when creating the Contentful client
+```js
+const contentful = require("contentful");
+const client = contentful.createClient({
+  accessToken: "<your-access-token>",
+  space: "<your-space-id>",
+  additionalTokens: {
+    "<other-space-id>": "<other-access-token>",
```
Cyberxon, line 76: maybe make it clear that the access token is that space access token by rephrasing to `<other-space-access-token>`

ADVANCED.md
```diff
@@ -62,6 +62,29 @@ The link resolution is applied to one level deep by default. If you need it to b
...
 
+### Links to other spaces (beta)
+
+`getEntry` and `getEntries` endponts can now resolve links to entries from other spaces as they do for regular links. The `resolveLinks`, `removeUnresolved` and `include` parameters affect both types of links.
+
+To use this functionality provide `additionalTokens` when creating the Contentful client
+```js
+const contentful = require("contentful");
+const client = contentful.createClient({
+  accessToken: "<your-access-token>",
```
Cyberxon, line 73: and here to `<current-space-access-token>`

Silhoue:
> Could you add a brief description to the PR indicating what this is adding and how it works?
Updated the PR description and added a section in the docs

luizfonseca (approved):
Besides the update on `index.d.ts`, no big comments from my side. 🚀

README.md
```diff
@@ -249,6 +249,13 @@ contentful.createClient({
         Set the environment that the API key has access to.
       </td>
     </tr>
+    <tr>
+      <td><code>additionalTokens</code></td>
+      <td><code>{}</code></td>
+      <td>
+        Additional access tokens, used for resolving links to other spaces. The expected format is <code>{ spaceId: accessToken, ... }</code>
```
Cyberxon, line 256: Also here in the readme maybe make it clear that each spaceId should have their respective access token

lib/contentful.js
```diff
@@ -75,20 +77,36 @@ export function createClient (params) {
...
-  })
-  // Append environment to baseURL
-  http.defaults.baseURL = getGlobalOptions().environmentBaseUrl
+    return createContentfulApi({
+      http: httpClient,
+      getGlobalOptions
+    })
+  }
+
+  const contentfulApiBySpaceId = Object.entries(config.additionalTokens).reduce((result, [spaceId, accessToken]) => {
```
Cyberxon, line 103: One improvement I can think of is comparing this access token to the access token provided for the main space and not calling the function if it is the case

Cyberxon (commented):
Great work! 💯  Just some minor comments

lib/contentful.js
```diff
@@ -75,20 +77,36 @@ export function createClient (params) {
...
+    return result
+  }, {})
+
+  contentfulApiBySpaceId[config.space] = createContentfulApiForSpace(config.space, config.accessToken, config.environment)
 
-  return createContentfulApi({
-    http,
-    getGlobalOptions
-  })
+  Object.values(contentfulApiBySpaceId).forEach(contentfulApi => { contentfulApi.extraApis = contentfulApiBySpaceId })
```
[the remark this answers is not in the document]
Silhoue: I renamed it to `additionalApis`, similar to `additionalTokens`.

PR #1675, opened 2023-02-09
fix(webpack): adjust bundler to avoid using function constructor

ruderngespra:
In order to avoid the usage of the `Function` operator in our browser bundle, as this runs into CSP issues (see https://github.com/webpack/webpack/issues/5627 and https://github.com/webpack/webpack/issues/6461) we adjust the webpack config: We set the `global` variable to `false` and by doing this avoid that webpack runs a check which makes use of the function constructor (https://github.com/webpack/webpack/blob/fb8afe71385734cfd65f47949117306a91f20753/buildin/global.js#L10).


Vinz93 (approved):
LGTM 👍

PR #1676, opened 2023-02-09
chore(es-check): check generated modules for es2017 instead of es6

ruderngespra:
As the generated modules are currently not satisfying the `es6` check anymore anyway (due to some overlooked js files in the `es-modules` root folder), we bump to `es2017` so what we are consistent among all artifact checks (browser and node are already only satisfying `es2017`). 

We also make sure that the js files in the root folder are included in the checks from now on (if there is a better blob pattern for this, please let me know).

This is where the checks were introduced: https://github.com/contentful/contentful.js/pull/390.


marcolink (approved):
LGTM

PR #1695, opened 2023-02-16
Simplify chain options

veu:
## Summary

Reduce chain options to one type by turning modifiers into a parameter.

## Description

So far every combination of chain options had its own type, leading to long and barely readable types. With a modifier parameter, types become slightly more readable.

E.g. `ChainOptionWithAllLocalesAndWithLinkResolutionAndWithUnresolvableLinks` vs `ChainOption<'allLocales'>`

The old chain options are still available to avoid conflicts.


lib/utils/client-helpers.ts
```diff
@@ -1,52 +1,47 @@
...
+  | 'noLinkResolution'
+  | 'noUnresolvableLinks'
+  | unknown
+  | undefined
+
+export type ChainOption<Modifiers extends ChainModifiers = undefined> = {
+  withoutLinkResolution: unknown extends Modifiers
+    ? boolean
+    : 'noLinkResolution' extends Modifiers
+    ? true
```
veu, line 12: Not sure what to say about this indentation. Looks like prettier is being creative. -.-

lib/create-contentful-api.ts
```diff
@@ -590,17 +582,17 @@ export default function createContentfulApi<OptionType extends ChainOptions>(
     Fields extends FieldsType,
     Locales extends LocaleCode,
     Options extends ChainOptions
-  > = Options extends ChainOptionWithAllLocalesAndWithLinkResolutionAndWithUnresolvableLinks
+  > = Options extends ChainOption<'allLocales'>
```
msieroslawska, line 585: Fot this piece of code and all similar type checking with ternaries: I don't know how the control flow works here and I'm wondering if it might be exiting where we don't want it to. I mean a situation like this:

```
const x, y;
if (x === 'foo' | y === 'bar') {
return something
} else if (x === 'foo') {
 return something else
}
return something something
```
In this example it will never enter `else if` because first `if` catches it first. The order of the flow should be changed in this case.

Is this a thing we should be careful with here?

Also since `Options extends ChainOption` is not last, does it mean the flow will never enter `Options extends ChainOption<'noUnresolvableLinks'>`?

veu: Fortunately, there is no way to skip ahead like that in the ternaries here. There’s no `return`, `break`, `continue` or similar. So we’re safe there.

> Also since `Options extends ChainOption` is not last, does it mean the flow will never enter `Options extends ChainOption<'noUnresolvableLinks'>`?

`ChainOption` is not a catch-all type, it’s the shorthand for `ChainOption<undefined>`. Thus `Options extends ChainOption` only evaluates to true if all options are `false` (the unconfigured/default client case). The order only matters if some types we check are sub types of each other.

msieroslawska (approved):

lib/utils/client-helpers.ts
```diff
@@ -1,52 +1,47 @@
+export type ChainModifiers =
+  | 'allLocales'
```
marcolink, line 2: If understood correctly, this is only used for discrimination, right? In this case, we could also think about using Capitalized Case strings, wdyt?

And is there a specific reason why not stick with the existing names ("including the "with")?

veu: We could also use capitalized modifiers. I’ll check separately what is best practice for these.

The main reason not to use the existing names was to keep it short. For the user interface the `with` version is nice to read but in a long list of conditional expressions with all combinations it helps with readability IMO.

PR #1696, opened 2023-02-16
Merge master into beta-resource-links

Silhoue:


veu (approved):

package.json
```diff
@@ -151,11 +151,20 @@
   "release": {
     "branches": [
       "master",
-      "next",
+      {
+        "name": "next",
```
Cyberxon, line 155: this next change is still experimental as part of my ticket, can we omit it from this bramch?

Silhoue: @Cyberxon Can we merge it now or is it still in progress?

Cyberxon: Thank you for the reminder, I just created a quick PR to fix that, feel free to approve and merge it so this will disappear

Cyberxon (approved):

PR #1700, opened 2023-02-17
chore: use simplified chain option types consistently

veu:
## Summary

Use new chain option types introduced in https://github.com/contentful/contentful.js/pull/1695 consistently and follow conventions for modifier names as suggested in  https://github.com/contentful/contentful.js/pull/1695#discussion_r1109549704.

## Description

* Replace remaining uses of old chain option types of format `ChainOptionWithAllLocalesAndWithoutLinkResolution`
* Move configured entry types to proper type file
* Update modifier enum names to match convention for constant names and match actual options / chain modifiers more closely.
* Make configured entry type more strict.

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -37,19 +37,19 @@ import {
   SyncCollection,
   Tag,
   TagCollection,
-  AbstractEntryCollection,
   Entry,
   ConfiguredAssetCollection,
   ConfiguredAsset,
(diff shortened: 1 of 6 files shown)


msieroslawska (approved):
This is much nicer to read! 🎉

marcolink (approved):
Without looking into every detail, the interface looks really readable 🎉

Cyberxon (approved):

PR #1705, opened 2023-02-20
fix: correct entry types and add chain modifiers

veu:
## Summary

Fix two issues with `Entry…` types and introduce a new configurable entry type to avoid similar issues in the future because we don’t have to write new definitions for every combination of modifiers.

### What’s in here
* Fix: `EntryWithoutLinkResolution` no longer allows resolved entries for reference fields.
* Fix: `EntryWithAllLocalesAndWithLinkResolutionAndWithUnresolvableLinks` now allows unresolvable links.
* Introduce `NewEntry` type that takes `ChainModifiers` to get the specific type.
* Introduce `GenericEntry` alias for `Entry` type and deprecate `Entry` type
* Introduce `GenericEntryCollection` alias for `EntryCollection` type and deprecate `EntryCollection` type.

### What’s not in here
The following changes will be done in a separate PR to reduce conflicts.
* Renaming the generic `Entry` type `GenericEntry`.
* Renaming the `NewEntry` type `Entry` and use it everywhere.
* Removing `ConfiguredEntry` in favor of `NewEntry`.
* Align entry collection naming with entry naming.

### DIFF
--- a/lib/types/entry.ts
+++ b/lib/types/entry.ts
@@ -42,21 +42,22 @@ export type BasicEntryField =
   | EntryFields.RichText
   | EntryFields.Object
 
+type BaseEntry = {
+  sys: EntrySys
+  metadata: Metadata
+}
+
 /**
  * @category Entities
  */
-export interface Entry<T> {
-  sys: EntrySys
-  metadata: Metadata
+export type Entry<T> = BaseEntry & {
   fields: T
 }
 
-export interface EntryWithAllLocalesAndWithoutLinkResolution<
+export type EntryWithAllLocalesAndWithoutLinkResolution<
   Fields extends FieldsType,
   Locales extends LocaleCode
-> {
-  sys: EntrySys
-  metadata: Metadata
+> = BaseEntry & {
   fields: {
     [FieldName in keyof Fields]: {
       [LocaleName in Locales]?: Fields[FieldName] extends EntryFields.Link<any>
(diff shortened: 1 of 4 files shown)


lib/types/entry.ts
```diff
@@ -42,109 +42,88 @@ export type BasicEntryField =
...
+  Modifiers extends ChainModifiers,
+  Locales extends LocaleCode = any
+> = Fields extends EntryFields.Link<infer LinkedFields>
+  ? 'WITHOUT_LINK_RESOLUTION' extends Modifiers
+    ? EntryLink
+    :
+        | NewEntry<LinkedFields, Modifiers, Locales>
+        | ('WITHOUT_UNRESOLVABLE_LINKS' extends Modifiers ? undefined : EntryLink)
+  : Fields extends EntryFields.Link<infer LinkedFields>[]
+  ? 'WITHOUT_LINK_RESOLUTION' extends Modifiers
```
veu, line 74: This check is similar to the one for single reference field and we could introduce a separate type that only has this check but then we’d have the check for each array element instead of once for the whole list and that didn’t feel cleaner. I’m open to suggestions though.

veu: I have a solution but it requires changing more than one type so it’ll be in the next PR.

lib/types/entry.ts
```diff
@@ -154,55 +133,49 @@ export interface AbstractEntryCollection<TEntry> extends ContentfulCollection<TE
...
-  : Options extends ChainOption<'WITHOUT_LINK_RESOLUTION'>
-  ? EntryWithoutLinkResolution<Fields>
-  : Options extends ChainOption<'WITH_ALL_LOCALES'>
-  ? EntryWithAllLocalesAndWithLinkResolutionAndWithUnresolvableLinks<Fields, Locales>
-  : Options extends ChainOption<'WITHOUT_LINK_RESOLUTION'>
-  ? EntryWithAllLocalesAndWithoutLinkResolution<Fields, Locales>
-  : Options extends ChainOption<'WITH_ALL_LOCALES' | 'WITHOUT_UNRESOLVABLE_LINKS'>
-  ? EntryWithAllLocalesAndWithLinkResolutionAndWithoutUnresolvableLinks<Fields, Locales>
-  : never
+> = Options extends ChainOption<infer Modifiers> ? NewEntry<Fields, Modifiers, Locales> : never
```
msieroslawska, line 178: I like this change 👍

Cyberxon (approved):
Beautifully done! The code is more readable now! Looking forward to the next iterations!

test/types/resolved-field.test-d.ts
```diff
@@ -0,0 +1,64 @@
...
+const entryValue = {
+  ...entryBasics,
+  fields: {
+    title: 'title',
+  },
+}
+
+type SimpleEntryFields = { title: string }
+
+expectAssignable<ResolvedField<Entry<SimpleEntryFields>, undefined>>(entryValue)
```
msieroslawska, line 44: You're deprecating `Entry` in this PR, shouldn't this be replaced with `NewEntry`? Or you're leaving it as it is since it will be renamed back in a follow up?

msieroslawska (approved):
Awesome work!

test/types/resolved-field.test-d.ts
veu: I don’t know what the formal process should look like actually. `NewEntry` can’t be renamed `Entry` until the current `Entry` is no longer used. I added a deprecation notice in the meantime to the old `Entry` to alert to that but yes, the name will technically stay.

PR #1711, opened 2023-02-22
fix: allow asset links in entry types

veu:
## Summary

Before this PR the entry types resolved all asset links to entries and entry links. Let’s fix that.

## Description

* Introduce `EntryFields.EntryLink` and `EntryFields.AssetLink` to distinguish between the different link types
* Allow resolved assets and asset links in `NewEntry` type
* Remove `FieldType` from generic entry types to indicate that the fields property can have different shapes
* Allow JSON data in `EntryFields.Object`
* Extract type tests mocks to separate file


test/types/asset-d.ts
```diff
@@ -12,89 +12,47 @@ import {
...
-  revision: numberValue,
-  space: { sys: { id: stringValue, type: 'Link', linkType: 'Space' } },
-  type: '', //?
-  updatedAt: dateValue,
-  id: stringValue,
-  createdAt: dateValue,
-}
-const metadataValue = { tags: [] }
+// @ts-ignore
+import * as mocks from './mocks'
```
veu, line 22: The error is that we’re only allowed to import from `lib` where we definitely don’t want the mocks. I’m doing a `*` import to make sure I’m not hiding other errors like mock data not being exported.

Cyberxon (approved):
Thank you for adding mocks and for the retyping effort! 💯 👍

PR #1716, opened 2023-02-23
chore: chain-specific behavior in parseEntries [DANTE-860]

Silhoue:
The `parseEntries` method no longer needs the config options, its input & return types are now determined via client chains, similar to `getEntry` and `getEntries`.

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -105,7 +105,7 @@ export type ClientWithAllLocalesAndWithoutLinkResolution = BaseClient &
     ): Promise<EntryCollection<Fields, 'WITH_ALL_LOCALES' | 'WITHOUT_LINK_RESOLUTION', Locales>>
 
     parseEntries<Fields extends FieldsType = FieldsType, Locales extends LocaleCode = LocaleCode>(
-      data: EntryCollection<Fields, 'WITH_ALL_LOCALES' | 'WITHOUT_LINK_RESOLUTION', Locales>
+      data: EntryCollection<Fields, 'WITHOUT_LINK_RESOLUTION' | 'WITH_ALL_LOCALES', Locales>
     ): EntryCollection<Fields, 'WITH_ALL_LOCALES' | 'WITHOUT_LINK_RESOLUTION', Locales>
   }
 
@@ -121,7 +121,7 @@ export type ClientWithLinkResolutionAndWithoutUnresolvableLinks = BaseClient &
     ): Promise<EntryCollection<Fields, 'WITHOUT_UNRESOLVABLE_LINKS'>>
 
     parseEntries<Fields extends FieldsType = FieldsType>(
-      data: EntryCollection<Fields, 'WITHOUT_LINK_RESOLUTION' | 'WITHOUT_UNRESOLVABLE_LINKS'>
+      data: EntryCollection<Fields, 'WITHOUT_LINK_RESOLUTION'>
     ): EntryCollection<Fields, 'WITHOUT_UNRESOLVABLE_LINKS'>
   }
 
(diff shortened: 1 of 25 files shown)


lib/types/entry.ts
```diff
@@ -175,6 +175,11 @@ export type GenericEntryCollection<Fields extends FieldsType> = AbstractEntryCol
   GenericEntry<Fields>
 >
 
+export type GenericEntryCollectionWithAllLocales<
```
veu, line 178: Is there a reason to introduce a new custom entry type?

lib/create-contentful-api.ts
```diff
@@ -84,34 +88,46 @@ export type ClientWithoutLinkResolution = BaseClient &
...
     ): Promise<EntryWithAllLocalesAndWithoutLinkResolution<Fields, Locales>>
 
-    getEntries<Fields extends FieldsType, Locales extends LocaleCode = any>(
+    getEntries<Fields extends FieldsType, Locales extends LocaleCode = LocaleCode>(
       query?: EntriesQueries<Fields> & { locale?: never }
     ): Promise<EntryCollectionWithAllLocalesAndWithoutLinkResolution<Fields, Locales>>
+
+    parseEntries<Fields extends FieldsType = FieldsType, Locales extends LocaleCode = LocaleCode>(
+      data: GenericEntryCollectionWithAllLocales<Fields, Locales>
+    ): EntryCollectionWithAllLocalesAndWithoutLinkResolution<Fields, Locales>
```
veu, line 130: The input and output types are incompatible. If the input data is allowed to have entries, then we can’t disallow them in the output as they will still be there.

Silhoue: Good point, we shouldn't allow resolved entries in the input. I changed the types.

lib/types/entry.ts
Silhoue: I removed it (but added different ones)

lib/create-contentful-api.ts
veu: The changed types allow anything as fields. They’re more generic than the last ones. If you want to disallow resolved entries with and without locales you should use these two:
`NewEntry<Fields, 'WITHOUT_LINK_RESOLUTION'>`
`NewEntry<Fields, 'WITH_ALL_LOCALES' | 'WITHOUT_LINK_RESOLUTION', Locales>`

test/unit/make-contentful-api-client-parseentries.test.ts
```diff
@@ -8,84 +39,77 @@ test('Given json should be parsed correctly as a collection of entries', () => {
...
     },
   }
-  const parsedData = api.parseEntries<any>(data)
+  const parsedData = api.parseEntries(data)
   expect(parsedData).toBeDefined()
-  expect(parsedData.items[0].fields.animal.sys).toEqual(data.includes.Animal[0].sys)
+  expect(parsedData.items[0].fields.animal?.sys).toEqual(pigEntry.sys)
 })
 
+// remove this test?
```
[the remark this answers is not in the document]
Silhoue: I don't see the value in having this test. Would it be fine to remove or should we keep it for some reason?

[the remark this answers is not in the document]
veu: [Here](https://github.com/contentful/contentful.js/pull/594#issuecomment-779251513)’s why the test was added. Does this still apply / seem useful?

test/integration/parseEntries.test.ts
```diff
@@ -0,0 +1,498 @@
...
+        details: { type: 'Link', linkType: 'Entry', id: '6SiPbntBPYYjnVHmipxJBF' },
+      },
+    ],
+    includes: {
+      Asset: [resolvedHappyCatAssetAllLocales],
+    },
+  }
+})
+
+// expand to cover also localeClient
```
veu, line 435: This is already done, right?

Silhoue: We're not using the `localeClient` but we have locales in the input data, so removed it.

test/unit/make-contentful-api-client-parseentries.test.ts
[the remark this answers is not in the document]
Silhoue: I found some more context on this ("How do we safely rollout tags in CDA CMA") so I guess it stays.

veu (approved):

PR #1722, opened 2023-02-27
fix: query filters by field type [DANTE-868]

Cyberxon:
## Summary

Checking all the query filters against their respective FieldTypes and making sure we have the same behaviour as the CDA.


lib/types/query/equality.ts
```diff
@@ -1,9 +1,9 @@
 import { BasicEntryField, EntryFields } from '../entry'
 import { ConditionalQueries } from './util'
 
-// TODO: Note: Equality and inequality operators are not supported for text fields
-// What types do we hav to exclude here?
-type SupportedTypes = Exclude<BasicEntryField, EntryFields.RichText> | undefined
+type SupportedTypes =
+  | Exclude<BasicEntryField, EntryFields.Object | EntryFields.RichText | EntryFields.Location>
```
veu, line 5: Are we sure this works? If one of the excluded types is a union (like `Object`) it means excluding all of the types in the union.

test/types/query-types/boolean.test-d.ts
```diff
@@ -47,3 +49,5 @@ expectAssignable<SubsetFilters<{ testField: EntryFields.Boolean }, 'fields'>>({
   'fields.testField[in]': booleanValue,
   'fields.testField[nin]': booleanValue,
 })
+
+// expectType<Required<FullTextSearchFilters<{ testField: EntryFields.Boolean }, 'fields'>>>({})
```
msieroslawska, line 53: Why is this commented out? Is it a todo?

Cyberxon: that was me testing sth out, thank you for noticing it!

msieroslawska (approved):
LGTM 🎈

PR #1733, opened 2023-03-02
chore: introduce generic client type

veu:
## Summary

Replace existing client types (`ClientWith…`) with one generic client type (`Client<ChainModifiers>`). With this change the client and entry types follow the same format which should make it easier to read and understand.

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -30,118 +30,75 @@ import { FieldsType } from './types/query/util'
 import normalizeSelect from './utils/normalize-select'
 import resolveCircular from './utils/resolve-circular'
 import validateTimestamp from './utils/validate-timestamp'
-import { ChainOption, ChainOptions } from './utils/client-helpers'
+import {
+  AddChainModifier,
+  ChainModifiers,
+  ChainOptions,
+  ModifiersFromOptions,
+} from './utils/client-helpers'
 import { validateLocaleParam, validateResolveLinksParam } from './utils/validate-params'
 
 const ASSET_KEY_MAX_LIFETIME = 48 * 60 * 60
 
-export type ClientWithLinkResolutionAndWithUnresolvableLinks = BaseClient &
-  BaseClientWithAssets & {
-    withAllLocales: ClientWithAllLocalesAndWithLinkResolutionAndWithUnresolvableLinks
-    withoutLinkResolution: ClientWithoutLinkResolution
-    withoutUnresolvableLinks: ClientWithLinkResolutionAndWithoutUnresolvableLinks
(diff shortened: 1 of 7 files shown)


lib/create-contentful-api.ts
```diff
@@ -30,118 +30,75 @@ import { FieldsType } from './types/query/util'
...
-    ): Promise<EntryCollection<Fields, undefined>>
-
-    // TODO: think about using collection generic as response type:
-    // ): Promise<Collection<EntryWithLinkResolution<Fields>>>
-
-    parseEntries<Fields extends FieldsType = FieldsType>(
-      data: EntryCollection<Fields, 'WITHOUT_LINK_RESOLUTION'>
-    ): EntryCollection<Fields, undefined>
-  }
+type ClientMethodsWithAllLocales<Modifiers extends ChainModifiers> = {
```
veu, line 43: Since I can’t dynamically add a generic for locales depending on whether the `withAllLocales` has been used, I kept two separate definitions for clients with and without `withAllLocales` internally but it’s only used here and not exported.

lib/create-contentful-api.ts
```diff
@@ -30,118 +30,75 @@ import { FieldsType } from './types/query/util'
...
 
-    parseEntries<Fields extends FieldsType = FieldsType, Locales extends LocaleCode = LocaleCode>(
-      data: EntryCollection<Fields, 'WITHOUT_LINK_RESOLUTION' | 'WITH_ALL_LOCALES', Locales>
-    ): EntryCollection<Fields, 'WITH_ALL_LOCALES' | 'WITHOUT_UNRESOLVABLE_LINKS', Locales>
-  }
+  getAssets(query?: AssetQueries<AssetFields>): Promise<AssetCollection<undefined>>
+}
 
-export type DefaultClient = ClientWithLinkResolutionAndWithUnresolvableLinks
+export type Client<Modifiers extends ChainModifiers> = BaseClient &
```
msieroslawska, line 86: Could you explain how this works? Especially why do we use `{}` when link resolution and unresolvable links modifiers are provided? Does it mean the returned type is just `BaseClient`?

veu: If the result is `{}` it means that we’re not adding methods in this section so the result will be a combination of `BaseClient` and the last part of the expression (with or without locales).
If you find a nicer way to do this I’m happy to clean this up. The only other way I know is setting the properties to `never` but that’s IMO even messier.

msieroslawska: Oh, I wish I knew a nicer way, IMO it's good as it is. Thanks for explaining!

msieroslawska (approved):
Looks great! 🎈

PR #1738, opened 2023-03-03
fix: limit select operator to allowed values

veu:
## Summary

Update select operator type to match the API.

## Description

* Split select filter into variations for assets and entries
* Require `content_type` when selecting for entry fields
* Remove `Partial` around `EntriesQueries` and `AssetQueries` because it makes required properties impossible
* Remove wildcard types from `EntriesQueries` and `AssetQueries` because it makes forbidden properties impossible
* Add simple `order` parameter type as stand in until we have a proper type for it

## Motivation and Context

We want to guide users more when using queries. Currently it’s only possible to get some autocomplete support but the types don’t give hints as to what’s required for certain operators or which properties are not allowed. This PR lays the groundwork to add more helpful query types.


msieroslawska (approved):
LGTM 🎈

Cyberxon (approved):
This looks so much better now! Thank you!

PR #1747, opened 2023-03-07
feat: require content type when it’s needed in API

veu:
## Summary

Require `content_type` in query for all operations on fields.

## Description

* Require `content_type` directly in `EntriesQueries` to keep types simple
* Remove defaults for the generic in `EntriesQueries` and `AssetQueries` as they’re always called with a value.
* Add tests for several operator types that focus on the `content_type` requirement.

### DIFF
--- a/lib/types/query/query.ts
+++ b/lib/types/query/query.ts
@@ -5,7 +5,7 @@ import { ExistenceFilter } from './existence'
 import { LocationSearchFilters } from './location'
 import { RangeFilters } from './range'
 import { FullTextSearchFilters } from './search'
-import { AssetSelectFilter, EntrySelectFilter } from './select'
+import { AssetSelectFilter, EntrySelectFilter, EntrySelectFilterWithFields } from './select'
 import { SubsetFilters } from './subset'
 import { FieldsType } from './util'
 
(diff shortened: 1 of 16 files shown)


lib/types/query/query.ts
```diff
@@ -26,26 +26,27 @@ export type SysQueries<Sys extends FieldsType> = ExistenceFilter<Sys, 'sys'> &
...
 
 // TODO: create-contentful-api complained about non-optional fields when initialized with {}
-export type EntriesQueries<Fields extends FieldsType = FieldsType> = EntryFieldsQueries<Fields> &
-  SysQueries<Pick<EntrySys, 'createdAt' | 'updatedAt' | 'revision' | 'id' | 'type'>> &
-  FixedQueryOptions &
-  FixedPagedOptions & { content_type?: string } & {
-    resolveLinks?: never
-  } & { order?: string }
+export type EntriesQueries<Fields extends FieldsType> =
+  | (EntryFieldsQueries<Fields> & { content_type: string })
```
msieroslawska, line 41: `content_type` should be allowed to be used as the only query as well, and I think line this only allows it to be paired with entry queries.

Cyberxon (approved):
now that I saw where the content type filter is, it seems more logical and readable!

lib/types/query/query.ts
Cyberxon: I think considering the fact most field queries are optional then that's included in this case

veu: All should be optional apart from `content_type`. This just says that you can’t use `EntryFieldsQueries` without `content_type`. If there is another property that is required, then we should make it optional.

veu: @msieroslawska There is one important detail that I learned recently that may not be obvious. If you have a type like `{ a?: string } | { b?: string }`, you’ll be able to combine the properties from the two parts of the union because `{}` matches both.  `{a: '', b: '' }` doesn’t look like it would match the type but it does. The behavior only changes when you have required properties.

PR #1765, opened 2023-03-14
fix: improve query types for getEntries

veu:
## Summary

Improve query types to better match API.

## Description

* add `links_to_asset` and `links_to_entry`
* enable equality, inequality, and full text search operators for symbol list fields


Cyberxon (approved):
LGTM

PR #1766, opened 2023-03-14
feat: add reference search to query types

veu:
## Summary

Add [search on references](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/search-parameters/search-on-references) to the query types. To not completely overload TypeScript with filters for nested entries, the specific paths are not checked in detail.


Cyberxon (approved):
LGTM 👍

PR #1767, opened 2023-03-14
fix: add proper type for TagQueries [DANTE-867]

Cyberxon:
## Summary

One of the TODOS is to properly type the tagQuery filters, this PR addresses that.

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -25,7 +25,7 @@ import {
   Entry,
   EntryCollection,
 } from './types'
-import { EntryQueries } from './types/query/query'
+import { EntryQueries, TagQueries } from './types/query/query'
 import { FieldsType } from './types/query/util'
 import normalizeSelect from './utils/normalize-select'
 import resolveCircular from './utils/resolve-circular'
@@ -234,8 +234,7 @@ interface BaseClient {
    * const response = await client.getTags()
    * console.log(response.items)
    */
-  // TODO type query
-  getTags(query?: any): Promise<TagCollection>
+  getTags(query?: TagQueries): Promise<TagCollection>
 
   /**
    * Creates an asset key for signing asset URLs (Embargoed Assets)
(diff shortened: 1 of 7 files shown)


lib/create-contentful-api.ts
```diff
@@ -234,8 +234,7 @@ interface BaseClient {
    * const response = await client.getTags()
    * console.log(response.items)
    */
-  // TODO type query
-  getTags(query?: any): Promise<TagCollection>
+  getTags(query?: TagQueries): Promise<TagCollection>
```
veu, line 237: We need some tests to check whether the tag query has the correct format.

Cyberxon: Sure I will add the tests

test/integration/getTags.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    console.dir(response, { depth: null })
+
+    expect(response.items[0].sys.type).toBe('Tag')
+  })
+
+  describe('tagName filters', () => {
+    it('gets the tag with the name equals to the provided value', async () => {
+      const response = await client.getTags({ name: 'public tag 1' })
+
+      expect(response.items.length).toBe(1)
```
msieroslawska, line 24: Nitpick: if you phrase the assertion to be `expect(response.items).toHaveLength(1)` it will have more human-friendly error message in case it fails ("expected 2 to be 1" vs "expected length: 1, received length: 2").

msieroslawska (approved):
LGTM, just small two comments 🎈

test/integration/getTags.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  params.insecure = true
+}
+
+const client = contentful.createClient(params)
+
+describe.only('getTags', () => {
+  it('returns all tags when no filters are available', async () => {
+    const response = await client.getTags()
+
+    console.dir(response, { depth: null })
```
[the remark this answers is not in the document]
Cyberxon: thank you for pointing it!

test/integration/getTags.test.ts
Cyberxon: Noted, I will start using that from now on

PR #1771, opened 2023-03-15
chore: chain-specific behavior in sync [DANTE-870]

msieroslawska:
The `sync` method no longer uses global options, but now accepts client chains of `withoutLinkResolution` and `withoutUnresolvableLinks` to get different return shape from the API depending on `resolveLinks` and `removeUnresolved`, similar to `getEntry` and `getEntries`. Tests and documentation have been updated to reflect the changes.

### DIFF
--- a/test/integration/tests.test.ts
+++ b/test/integration/tests.test.ts
@@ -1,7 +1,7 @@
 import { EntryFields } from '../../lib'
 import * as contentful from '../../lib/contentful'
 import { ValidationError } from '../../lib/utils/validation-error'
-// @ts-ignore
+import { TypeCatFields } from './parseEntries.test'
 import { localeSpaceParams, params, previewParams } from './utils'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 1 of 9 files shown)


test/integration/tests.test.ts
Cyberxon, line 1: I prefer we add the sync tests in their own file because this `tests.test.ts` file has too many tests already and is becoming more and more unmaintainable.

test/unit/paged-sync.test.ts
```diff
@@ -48,6 +53,7 @@ describe('paged-sync', () => {
         {
           initial: true,
           content_type: 'id',
+          // @ts-ignore
```
Cyberxon, line 56: why was this `@ts-ignore` added?

Cyberxon (approved):
LGTM, just minor comments.

test/unit/paged-sync.test.ts
msieroslawska: It's a test which checks if request is rejected for invalid content type and type. New types force you to only use `Entry` as type if `content_type` is provided, therefore I need to force TS to ignore the invalid value to run the test.

Cyberxon: Oh I understand it more now, thank you!

PR #1774, opened 2023-03-16
fix: remove the locale query filter from the list of FixedQueryOptions [DANTE-867]

Cyberxon:
## Summary

This PR addresses the second TODO about omit the locale from the list of fixed query options and only adding it where we don't have `WithAllLocales` chain modifier.

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -25,8 +25,8 @@ import {
   Entry,
   EntryCollection,
 } from './types'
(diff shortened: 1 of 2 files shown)


msieroslawska (approved):
LGTM 🎈 
I am wondering though, why are there no changes needed in tests? Since we use the locale option only where it is necessary, shouldn't it mean that some tests are affected? Or do we miss coverage in this area?

PR #1783, opened 2023-03-20
chore: deprecate old client config

veu:
## Summary

Remove `resolveLinks` and `removeUnresolved` completely after they’ve been deprecated in all relevant methods.

## Description

* Remove defunct options from `createClient`
* Remove related tests and move `getEntries` tests into one file for consistency
* Update v10 migration information
* Reduce _Response configuration options_ section to single line

PR #1785, opened 2023-03-20
chore: throw error when using old parameters

veu:
## Summary

Follow-up for https://github.com/contentful/contentful.js/pull/1783 to remove the old `resolveLinks` and `removeUnresolved` parameters on the JS level.

## Description

* Throw error when `resolveLinks` or `removeUnresolved` are used in query to alert JS users to new API.
* Throw error when `resolveLinks` or `removeUnresolved` are used in client creation to alert JS users to new API.
* Remove old parameters from global options.

### DIFF
--- a/lib/contentful.ts
+++ b/lib/contentful.ts
@@ -9,6 +9,7 @@ import { ContentfulClientApi } from './create-contentful-api'
 import createGlobalOptions from './create-global-options'
 import { makeClient } from './make-client'
 import type { AxiosRequestConfig } from 'axios'
+import { validateRemoveUnresolvedParam, validateResolveLinksParam } from './utils/validate-params'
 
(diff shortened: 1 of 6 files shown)


lib/contentful.ts
```diff
@@ -41,6 +42,9 @@ export function createClient(params: CreateClientParams): ContentfulClientApi {
     throw new TypeError('Expected parameter space')
   }
 
+  validateResolveLinksParam(params)
```
msieroslawska, line 45: Since chain options are only to be used on the client instance, not directly when creating it, I'm wondering if error messages for these validators are clear enough to be used here too (so users don't try anything like `contentful.withoutLinkResolution.createClient(params)`). What do you think?

msieroslawska (approved):

PR #1789, opened 2023-03-21
feat: add order query parameter

veu:
## Summary

Add `order` parameter for `getAssets`, `getEntries`, and `getTags` queries.

### DIFF
--- a/lib/types/query/order.ts
+++ b/lib/types/query/order.ts
@@ -1,0 +1,68 @@
+import { FieldsType } from './util'
+import { EntryFields, EntrySys } from '../entry'
+import { AssetSys } from '../asset'
+import { ConditionalPick } from 'type-fest'
+import { TagSys } from '../tag'
+
+type SupportedTypes =
+  | EntryFields.Symbol
+  | EntryFields.Integer
+  | EntryFields.Number
+  | EntryFields.Date
+  | EntryFields.Boolean
+  | undefined
+
+type SupportedLinkTypes = EntryFields.AssetLink | EntryFields.EntryLink<any> | undefined
+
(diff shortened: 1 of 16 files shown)


lib/types/query/order.ts
```diff
@@ -0,0 +1,68 @@
...
+  | EntryFields.Integer
+  | EntryFields.Number
+  | EntryFields.Date
+  | EntryFields.Boolean
+  | undefined
+
+type SupportedLinkTypes = EntryFields.AssetLink | EntryFields.EntryLink<any> | undefined
+
+export type OrderFilterPaths<Fields extends FieldsType, Prefix extends string> =
+  | `${Prefix}.${keyof ConditionalPick<Fields, SupportedTypes> & string}`
```
Silhoue, line 18: Why do we need the `& string` here? Isn't `keyof FieldsType` always a string?

test/types/queries/asset-queries.test-d.ts
```diff
@@ -123,6 +123,21 @@ expectNotAssignable<DefaultAssetQueries>({
   'sys.unknownProp[nin]': mocks.anyValue,
 })
 
+// order operator
+
+expectAssignable<DefaultAssetQueries>({ order: ['sys.createdAt'] })
```
Silhoue, line 128:
```suggestion
expectAssignable<DefaultAssetQueries>({ order: ['sys.createdAt', '-sys.createdAt'] })
```

Silhoue (approved):

lib/types/query/order.ts
veu: In some cases it’s just string. Here we get `string | number` so we have to remove `number` to make TS happy.

PR #1796, opened 2023-03-23
feat: add EntrySkeletonType combining fields with content type id [DANTE-946]

Silhoue:
Introduce `EntrySkeletonType` type that replaces `FieldsType` in calls to `getEntry`, `getEntries`, and `parseEntries`. In addition to fields it contains the content type id and allows stronger typing of the returned values.

### DIFF
--- a/TYPESCRIPT.md
+++ b/TYPESCRIPT.md
@@ -78,7 +78,7 @@ Static query keys are not influenced by the shape of the entries or assets you'r
 
 ![](images/static-query-keys.png)
 
-```
+```js
 getEntries({
     'skip': 10,
     'limit': 20,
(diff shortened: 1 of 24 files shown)


TYPESCRIPT.md
```diff
@@ -92,23 +92,26 @@ Dynamic query keys are based on the given shape of the expected entries' content
 
 ![](images/dynamic-query-keys.png)
 
-To calculate dynamic keys, we have to define the shape of the fields of the entries' content type:
+To calculate dynamic keys, we have to define the shape the entries' content type:
```
veu, line 95: Combining "shape" and "content type" is somewhat duplicated. Either "we have to provide the shape of the entries:" or "we have to provide the relevant information from the content type:" would work for me.

veu (approved):
Exciting!

msieroslawska (approved):

PR #1823, opened 2023-03-29
chore: updating markdown files with relevant information

Cyberxon:
## Description

Updating markdown documentation with the latest information of the typescript project


ADVANCED.md
```diff
@@ -122,20 +122,18 @@ Please see the notes below for link resolution prior to v.10.0.0 and v.7.0.0.
 
 #### Note: link resolution for versions older than 10.0.0
 
-Please note that for versions older than 10.0.0, disabling link resolution needs to be done via [configuration options](README.md#response-configuration-options) during client creation.
-To disable it, set `resolveLinks` to `false` when creating the Contentful client. Like so:
+Please note that for versions older than 10.0.0, disabling link resolution needs to be done via [Chained Clients](README.md#chained-clients).
```
veu, line 125: No, chained clients are for v10. This didn’t change.

TYPESCRIPT.md
```diff
@@ -116,10 +117,34 @@ getEntries<ExampleEntrySkeleton>({
 })
 ```
 
-#### Limitation
+#### Limitations
 
-Due to the limitation with recursive types, we can only calculate keys on the root level of your content type.
+- Due to the limitation with recursive types, we can only calculate keys on the root level of your content type.
 There is currently no way to calculate keys for nested (recursive) content types.
+- To limit the complexity of query types we use a simple type definition for [search on references](https://contentful.atlassian.net/wiki/spaces/PROD/pages/4169466000/Known+limitations+of+contentful.js+TypeScript+support#:~:text=search%20on%20references).
```
veu, line 124: This refers to the same limitation as the first point. Let’s remove the first as that is less accurate (it’s possible to calculate the keys, just not desirable if you want to actually use the types).

Cyberxon: I was aware that they were very similar just needed your input on which one is better suited, thank you!

ADVANCED.md
Cyberxon: oh, nice catch

veu (approved):

PR #1829, opened 2023-03-30
fix: examples in TYPESCRIPT.md

Cyberxon:
## Summary

A quick fix for the TYPESCRIPT.md examples

### DIFF
--- a/TYPESCRIPT.md
+++ b/TYPESCRIPT.md
@@ -94,17 +94,20 @@ Dynamic query keys are based on the given shape of the expected entries' content
 
 To calculate dynamic keys, we have to provide the shape of the entries:
 
+:warning: Please use `contentful.EntryFieldTypes` instead of `contentful.EntryFields` when defining your entry squeleton types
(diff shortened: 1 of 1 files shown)


veu (changes requested):
Let’s fix the typos and try again.

TYPESCRIPT.md
```diff
@@ -220,12 +238,22 @@ const client = contentful.createClient({
...
+  }
+}
+
+type ReferencedProductSkeleton = {
+  fields: { relatedProduct: contentful.EntryFieldTypes.EntryLink<ProductSqueleton> },
+  contentTypeId: 'referencedProduct'
 }
 type Locales = 'en-US' | 'de-DE'
-const entry = client.withoutLinkResolution.getEntry<ProductSkeleton, Locales>('some-entry-id')
+// Chaining withAllLocales to be able to add the Locale type
```
veu, line 255: I don’t think we need the comment here as we have examples for `withAllLocales` where this is already explained. If it’s becoming too complicated we can also remove the locales here.

PR #1832, opened 2023-03-30
feat: add documentation for simple type definitions

msieroslawska:
I have added more type descriptions and documentation links for simple types.

### DIFF
--- a/lib/types/content-type.ts
+++ b/lib/types/content-type.ts
@@ -4,7 +4,9 @@ import { SpaceLink, EnvironmentLink } from './link'
 import { BaseSys } from './sys'
 
 /**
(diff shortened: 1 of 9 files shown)


lib/types/content-type.ts
```diff
@@ -79,6 +85,7 @@ export interface ContentTypeFieldValidation {
 }
 
 /**
+ * @TODO
```
Cyberxon, line 88: why is this labeled TODO?

Cyberxon (approved):
LGTM, aside from one comment

lib/types/content-type.ts
veu: Because it’s not documented yet.

PR #1859, opened 2023-04-06
fix: correct link for build status in all .md files

Cyberxon:
## Summary

fix the Build status link and update the breaking change section of the MIGRATION.md to include the query type breaking change


marcolink (approved):
merci!

PR #1870, opened 2023-04-11
fix: revert axios version

Cyberxon:
## Summary

reverting the axios version that was causing tests and pipeline to fail


veu (approved):

ruderngespra (approved):
lgtm, I assume this same issue exists on the master branch and isn't fixed there either?

PR #1872, opened 2023-04-11
chore: update typedoc docs main page

veu:
## Summary

Small update for the  TypeDoc main page.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -4,9 +4,9 @@
   </a>
 </p>
 
-<h1 align='center'>Content Delivery API</h1>
+<h1 align='center'>contentful.js</h1>
 
-<h3 align="center">TypeScript type documentation</h3>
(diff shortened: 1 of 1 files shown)


DOCS.md
msieroslawska, line 1: Headers and intro have been taken from existing `README.md`. If we want to change the wording, I'd suggest updating the other file as well.

veu: I can also revert those changes if that suits you better. The important parts are from line 34 onwards.

msieroslawska: It's not about me :) The wording changes are nothing major, I only assumed that we'd want to be consistent across this readme-ish doc intro and the actual README.

msieroslawska (approved):

DOCS.md
veu: "TypeScript type documentation" was never mentioned in the README and the new version is more consistent with the v9 docs to avoid surprises there.

PR #1879, opened 2023-04-13
feat: support in validation in entry field types

veu:
## Summary

Add support for the `in` validation in field types allowing users to narrow down number and text field values.

## Description

For example, take a content type field with the following validation.

```
{
  "id": "textField",
  "name": "text field",
  "type": "Symbol",
  "validations": [
    {
      "in": ['option 1', 'option 2']
    }
  ],
  …
}
```

We can translate this to the following types.

```typescript
type ExampleEntrySkeleton = EntrySkeletonType<{
  textField: EntryFieldTypes.Symbol<'option 1' | 'option 2'>,
}>

type ExampleEntry = Entry<ExampleEntrySkeleton>
```

## Motivation and Context

In contentful.js v9 this was already possible because we used native types to describe fields. Since we’re using abstract types in v10 we have to explicitly allow this.

### DIFF
--- a/lib/types/entry.ts
+++ b/lib/types/entry.ts
@@ -25,10 +25,10 @@ export interface EntrySys extends EntitySys {
  * @category Entry
  */
 export declare namespace EntryFieldTypes {
-  type Symbol = { type: 'Symbol' }
-  type Text = { type: 'Text' }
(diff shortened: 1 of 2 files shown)


Cyberxon (approved):
LGTM! Great work!

Issue #1901, opened 2023-05-04
SyntaxError Unexpected token '?' in file create-contentful-api.ts


wadehammes:
## Expected Behavior

No error should occur.

## Actual Behavior

Getting a SyntaxError Unexpected token '?' in file create-contentful-api.ts, which appears to only be happening on older browser versions, and has been happening ever since 10.0.0 was released. It has been causing "Application error: a client-side error has occurred" within my Next.js application the last few weeks for only a small subset of users using outdated browser versions.

<img width="543" alt="image" src="https://user-images.githubusercontent.com/222170/236294288-b790c99e-d460-468b-af77-2a30ca1b1f8b.png">

## Possible Solution

Not sure, from what I gather this is a compiling issue potentially?

## Steps to Reproduce

Have next.js installed with latest Contentful, use a old browser like Chrome 75 or something.

wadehammes:
Anything regarding this?

marcolink:
Hi @wadehammes,

Would you be able to provide more details about the browser where this occurs? [Here](https://github.com/contentful/browserslist-config/blob/main/index.js) you can find the `browserlist` config we use to build our libraries.

wadehammes:
It was happening in various browsers, mostly chrome. I saw one instance of it in Chrome 104. Most were mobile Chrome versions going back to 75.

wadehammes:
I also want to add that reverting back to 9.3.5 resolved the error and I haven't seen it again in Sentry, so definitely isolated to this update.

wadehammes:
@marcolink is this something your team can look into, it's preventing me from updating. I am happy to provide more sentry logs or whatever is needed.

dsolanorush:
I am also seeing this error, along with a couple of other errors from the same file (create-contentful-api.ts), including `Unexpected token ';'. Expected an opening '(' before a method's parameter list.`. I also reverted back to `9.3.5` to resolve the issue

I would certainly like to upgrade if this issue can be identified and resolved. Any progress @marcolink?

wadehammes:
Any movement here @marcolink ?

marcolink:
@wadehammes 
a new version with higher browser compatibility has just been released https://github.com/contentful/contentful.js/releases/tag/v10.3.1

Please let us know if this is still not solving your case!

wadehammes:
[...]

PR #1948, opened 2023-06-27
fix: downgrade browseslist to support old browsers [PHX-2717]

marcolink:
## Summary

Customers reported issues with javascript language features (optional chaining) on old browsers. 
By Downgrading the min supported browserslist config, we expect the build to polyfill unknown language features. 

The version is selected based on [this list](https://caniuse.com/?search=optional%20chaining).

Solves: #1901

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,10 @@
   "browser": "./dist/contentful.browser.min.js",
   "types": "./dist/types/index.d.ts",
   "browserslist": [
-    "extends @contentful/browserslist-config"
+    "Chrome >= 79",
+    "Edge >= 79",
+    "Firefox >= 73",
+    "Safari >= 13"
   ],
   "tsd": {
     "directory": "test/types",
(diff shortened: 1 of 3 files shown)


veu (approved):

package.json
```diff
@@ -7,7 +7,10 @@
   "browser": "./dist/contentful.browser.min.js",
   "types": "./dist/types/index.d.ts",
   "browserslist": [
-    "extends @contentful/browserslist-config"
+    "Chrome >= 79",
```
Vinz93, line 10: I think we don't want the version `79` right? 🤔

veu: The one they mentioned was 75 so even lower. This isn’t about recommending good practices though. Just increasing compatibility for those who need it.

wadehammes:
Not sure if you saw my comment on the issue that I created for this, but this did not solve the issue unfortunately. https://github.com/contentful/contentful.js/issues/1901

PR #1949, opened 2023-06-28
fix: allow null type for validation message

BasKiers:
## Summary

Allow `null` value for validation `message`

## Description

It turns out the `message` value for validations can be set to `null`. This PR updates the Typescript type to reflect this.

## Motivation and Context

The Typescript type is not correct for the possible value set for `message`


andreascful (approved):
Thank you. Looks great

PR #2012, opened 2023-09-15
fix: reintroduce babel to ensure compatiblitiy with browserlist

ruderngespra:


marcolink:
@ruderngespra should this be merged?

PR #2025, opened 2023-09-28
feat: add metadata types for select query [ZEND-4163]

mayakarabula:
For select query we are missing types for metadata, thus causing errors for users trying to use it


Vinz93 (approved):
LGTM 🙂

marcolink:
I think we still had an open question on "MixedResults" (no content type id provided, but filtered by tags). But I currently don't see what should block this change 👍

marcolink (approved):

PR #2085, opened 2023-11-29
chore: adjust bundle size limits

ruderngespra:
Unfortunately we have to adjust the bundle size expectations a bit to enable further version bumps from `contentful-sdk-core`. This is just to unblock updates, on the long run we might look into getting rid of additional dependencies if this becomes too big.

[Blocked PR](https://github.com/contentful/contentful.js/pull/2084)

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,19 +129,19 @@
   "bundlesize": [
     {
       "path": "./dist/contentful.node.js",
-      "maxSize": "105Kb"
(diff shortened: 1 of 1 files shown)


ruderngespra:
cc @contentful/team-developer-experience Are okay with this change?

mgoudy91 (approved):
seems reasonable to me!

PR #2209, opened 2024-04-25
feat: extend SDK with content source maps support [TOL-2043]

2wce:
## Summary

Content source maps will change the sys object in the entity for the response and this has been implemented as follows

- [x] When creating the Client for the SDK, we have added an option to include alpha features where csm can be enabled/disabled at client level. 
- [x]  We have added a wrapper function which will act like middleware of sorts & check if the alpha feature is enabled before adding the includeContentSourceMaps param to the query to enable CSM
- [x] We have added tests to ensure that the alpha feature can be correctly enabled
- [x] we have also added tests to show that adding Content Source Maps won't break any existing logic for fetching assets & entries including the current supported chain modifers 

## Motivation and Context

This has been added as part of this [CEP](https://contentful.atlassian.net/wiki/spaces/ENG/pages/4538695717/CEP-0195+Content+Source+Maps+in+the+REST+API+CPA) and this [ticket](https://contentful.atlassian.net/browse/TOL-2043?atlOrigin=eyJpIjoiNDk3NWQ4ZTA4NTNkNDI1ZDhlZTU3NGJlYjJkOTI1MWQiLCJwIjoiaiJ9) in order to let users fetch Content Source Maps with the REST API

### DIFF
--- a/lib/contentful.ts
+++ b/lib/contentful.ts
@@ -114,6 +114,11 @@ export interface CreateClientParams {
    * Enable alpha features.
    */
   alphaFeatures?: {
+    /**
+     * Enable Content Source Maps.
+     * @remarks
+     * This feature is only available when using the Content Preview API.
+     */
     withContentSourceMaps?: boolean
   }
 }
--- a/lib/contentful.ts
+++ b/lib/contentful.ts
@@ -109,6 +109,13 @@ export interface CreateClientParams {
    * Interceptor called on every response. Takes Axios response object as an arg.
    */
   responseLogger?: (response: AxiosResponse<any> | Error) => unknown
+
+  /**
+   * Enable alpha features.
+   */
+  alphaFeatures?: {
+    withContentSourceMaps?: boolean
+  }
 }
 
 /**
--- a/lib/contentful.ts
+++ b/lib/contentful.ts
@@ -3,13 +3,13 @@
  * with access to the Contentful Content Delivery API.
  */
 
+import type { AxiosAdapter, AxiosRequestConfig, AxiosResponse } from 'axios'
 import axios from 'axios'
 import { createHttpClient, getUserAgentHeader } from 'contentful-sdk-core'
 import { createGlobalOptions } from './create-global-options'
 import { makeClient } from './make-client'
-import type { AxiosAdapter, AxiosRequestConfig, AxiosResponse } from 'axios'
-import { validateRemoveUnresolvedParam, validateResolveLinksParam } from './utils/validate-params'
 import { ContentfulClientApi } from './types'
+import { validateRemoveUnresolvedParam, validateResolveLinksParam } from './utils/validate-params'
 
 /**
  * @category Client
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -98,10 +98,10 @@ export default function createContentfulApi<OptionType extends ChainOptions>(
     return baseUrl
   }
 
-  const alphaFeatures = (http.httpClientParams as any as CreateClientParams)?.alphaFeatures
-  const host = http.httpClientParams.host ?? new URL(http.defaults.baseURL as string).host
-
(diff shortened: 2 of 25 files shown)


lib/contentful.ts
```diff
@@ -109,6 +109,13 @@ export interface CreateClientParams {
    * Interceptor called on every response. Takes Axios response object as an arg.
    */
   responseLogger?: (response: AxiosResponse<any> | Error) => unknown
+
+  /**
+   * Enable alpha features.
+   */
+  alphaFeatures?: {
+    withContentSourceMaps?: boolean
```
z0al, line 117: It would be helpful If you add a JS Doc here explaining briefly what is this about for customers. It will show up to the developer as a help text in the editor completion menu.

lib/create-contentful-api.ts
```diff
@@ -96,6 +97,15 @@ export default function createContentfulApi<OptionType extends ChainOptions>(
     return baseUrl
   }
 
+  const alphaFeatures = (http.httpClientParams as any as CreateClientParams)?.alphaFeatures
+
+  function maybeEnableSourceMaps(query: Record<string, any> = {}): Record<string, any> {
+    if (alphaFeatures?.withContentSourceMaps) {
+      query.includeContentSourceMaps = true
```
z0al, line 104: We should validate against enabling Content Source Maps for CDA and report a user friendly error.

test/integration/getEntries.test.ts
```diff
@@ -410,7 +458,21 @@ function assertLocalizedEntriesResponse(response) {
...
+
   expect(response.items[0].fields.bestFriend['en-US'].fields).toBeDefined()
   expect(response.items[0].fields.bestFriend['en-US'].sys.type).toBe('Entry')
   expect(response.items[0].metadata).toEqual({ tags: [] })
 }
+
+function assertCSMEntriesResponse(response) {
+  expect(response.includes).toBeDefined()
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  expect(response.includes!.Asset).toBeDefined()
```
z0al, line 470: Nit: Are these non-null assertions really needed? Why not use `response.includes?.Asset` ?

test/integration/getEntries.test.ts
```diff
@@ -410,7 +458,21 @@ function assertLocalizedEntriesResponse(response) {
...
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  expect(Object.keys(response.includes!.Asset!).length).toBeGreaterThan(0)
+
+  expect(response.items[0].fields.bestFriend.fields).toBeDefined()
```
z0al, line 474: Nit: can we set `response.items[0]` to a variable and use that?

test/unit/contentful.test.ts
```diff
@@ -116,4 +116,16 @@ describe('contentful', () => {
       'http://some-base-url.com/environments/stage',
     )
   })
+
+  test('Initializes API with alpha features', () => {
```
z0al, line 120: Not sure how useful this test is atm. 

What I'm missing so far in other tests is that none of explicitly checks if we are passing the correct query param. 

That's more useful than just checking if we are passing the `alphaFeatures` option down the chain.

z0al (commented):
Looks great overall. Just a few pieces to iron out.

lib/create-contentful-api.ts
aodhagan-cf: could the `alphaFeatures?.withContentSourceMaps` part be done in the `checkIncludeContentSourceMapsParamIsAllowed` function? Then you could simply check isAllowed

nit: `areAllowed`

2wce: sounds good 👌🏽i had picked `isAllowed` since we are checking if the param is allowed not necessarily if content source maps are allowed but it's fine either way I'll push the change

aodhagan-cf: Ah right makes sense then sorry, yeah also happy with either

jjolton-contentful (approved):

PR #2218, opened 2024-05-07
feat: encode CPA response when csm alpha feature is enabled [TOL-2080]

2wce:
## Summary

This should make it easier for customers to set up Content Source Maps. Once this is merged, enabling the CSM alpha feature will encode the CPA response directly reducing the steps needed to set this up

## Description

We have added a wrapper function the conditionally encodes the CPA response depending on whether the alpha feature is enabled or not

## Todos

-   [x] rename the alpha feature from `withContentSourceMaps` to `includeContentSourceMaps`
-   [x] it should encode the CPA response directly (encodeCPAResponse)
-   [x] it loads the content source maps in the background
-   [x] Import it from the content-source-maps package (smaller bundlesize than live preview SDK)

NOTE:
We haven't added tests for the entry endpoint because the test entry in the space uses unsupported fields & we have realised that we currently encode them in the csm package anyway so we would have to finish the work on supporting them first before adding tests here to confirm if they work

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -3,6 +3,7 @@
  * different kinds of entities present in Contentful (Entries, Assets, etc).
  */
 
+import { encodeCPAResponse } from '@contentful/content-source-maps'
 import { AxiosInstance, createRequestConfig, errorHandler } from 'contentful-sdk-core'
 import { CreateClientParams } from './contentful'
 import { GetGlobalOptions } from './create-global-options'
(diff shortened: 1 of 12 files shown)


lib/create-contentful-api.ts
```diff
@@ -122,12 +123,22 @@ export default function createContentfulApi<OptionType extends ChainOptions>(
...
+    return data
+  }
+
   async function get<T>({ context, path, config }: GetConfig): Promise<T> {
     const baseUrl = getBaseUrl(context)
 
     try {
       const response = await http.get(baseUrl + path, config)
-      return response.data
+      return maybeEncodeCPAResponse(response.data, config)
```
z0al, line 141: Are we sure this only applies to responses from entries and assets endpoints?

2wce: yes `includeContentSourceMaps` is only set to true when we use maybeEnableSourceMaps like [this](https://github.com/contentful/contentful.js/blob/980b087f5bd0381eb6b1c289b53e55c37d049f1e/lib/create-contentful-api.ts#L260) & we have only added it to the entries and assets endpoints

package.json
```diff
@@ -174,5 +177,8 @@
       "@semantic-release/changelog",
       "@semantic-release/github"
     ]
+  },
+  "relativeDependencies": {
```
YvesRijckaert, line 181: I think this should be removed :)

z0al (dismissed):
LGTM but I'm missing tests for the new behavior.

mgoudy91 (approved):
looks good to me, no concerns

PR #2219, opened 2024-05-08
feat: ensure entity.sys is selected when withContentSourceMaps is enabled

z0al:
## Summary

Ensure that `sys` is selected when `withContentSourceMaps` option is enabled but `sys` isn't present in `query.select`

### DIFF
--- a/lib/utils/query-selection-set.ts
+++ b/lib/utils/query-selection-set.ts
@@ -1,0 +1,14 @@
+export default function getQuerySelectionSet(query: Record<string, any>): Set<string> {
+  if (!query.select) {
+    return new Set()
+  }
+
+  // The selection of fields for the query is limited
+  // Get the different parts that are listed for selection
+  const allSelects = Array.isArray(query.select)
+    ? query.select
+    : query.select.split(',').map((q) => q.trim())
+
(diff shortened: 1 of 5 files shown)


2wce (approved):
lgtm 🚀🚀

lib/utils/query-selection-set.ts
```diff
@@ -0,0 +1,14 @@
+export default function getQuerySelectionSet(query: Record<string, any>): Set<string> {
+  if (!query.select) {
+    return new Set()
+  }
+
+  // The selection of fields for the query is limited
+  // Get the different parts that are listed for selection
+  const allSelects = Array.isArray(query.select)
+    ? query.select
+    : query.select.split(',').map((q) => q.trim())
```
aodhagan-cf, line 10: nit: maybe `q` here could be `parameter` or even `param`?

aodhagan-cf: actually not sure if that's the best name, I'm just not a fan of single character variables. Just in case in future the function gets bigger and context can get lost

aodhagan-cf (approved):

mgoudy91 (approved):

Issue #2246, opened 2024-06-06
Retried requests duplicate  query parameters resulting in 400 response


geigerzaehler:
When a request is retried query parameters are duplicated which results in a 400 response.

This is a bit tricky to reproduce, but possible with Chrome Dev Tools’ [network request blocking](https://developer.chrome.com/docs/devtools/network-request-blocking) feature. I have a simple script that I load in the browser:

```
await client.getEntries({ contentType: "foo" })
```

In the dev tools I add “contentful“ to the list of blocked requests.
![image](https://github.com/contentful/contentful.js/assets/3919579/dae53f9b-1c21-4829-a6f9-e1b71e5f992f)

Then I reload the page. In the network tab of the dev tools I can see that requests to contentful are blocked. I quickly uncheck contentful in the request blocking tab to let it go through. But the server responds with a 400 and the message “The query you sent included duplicate URI query parameters...”.

![image](https://github.com/contentful/contentful.js/assets/3919579/fcd1ed7f-783b-47ad-b292-131584e06405)

From the network tab I can see that the request URL looks like this:
```
https://cdn.contentful.com/spaces/m2xydvefjw31/environments/master/entries?content_type=foo&content_type=foo&content_type=foo
```

It looks like every retry appends the query string again.

This issue started to occur in 10.11.1 and wasn’t present in 10.11.0.

Techn1x:
We're experiencing the same issue from the same version and have put through a support request to contentful

t-col:
Hi @geigerzaehler thanks for reporting. A stack trace of the `The query you sent included duplicate URI query parameters` error would be helpful in diagnosing where this problem is occurring. I have not been able to reproduce yet with the same setup you're describing.

geigerzaehler:
Thanks for talking a look @t-col. Here’s the stack trace:
```
BadRequest
    at Ur (http://localhost:3000/node_modules/.vite/deps/contentful.js?v=9eaebbc3:6774:22)
    at a2 (http://localhost:3000/node_modules/.vite/deps/contentful.js?v=9eaebbc3:7217:17)
    at async u2 (http://localhost:3000/node_modules/.vite/deps/contentful.js?v=9eaebbc3:7224:27)
   [REDACTED]
```

It looks like the package does not provide any sourcemaps so this doesn’t look helpful unfortunately.

Can I make the steps to reproduce the issue clearer in any way? Which part can’t you reproduce?

Techn1x:
[...]

PR #2253, opened 2024-06-12
fix: pin axios to 1.6.x to prevent request params from being duplicated on retries

t-col:
## Summary

addresses https://github.com/contentful/contentful.js/issues/2246

The bug in our SDK can be traced to the upgrade of `axios` from 1.6.8 to 1.7.x ([here](https://github.com/contentful/contentful.js/compare/v10.11.0...v10.11.1)). I'm not yet certain if this new version of `axios` has a bug, of if a fix there exposed a bug in our implementation of it. For now we'll pin our `axios` dependency to the last known compatible minor version, 1.6.x

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@contentful/content-source-maps": "^0.5.0",
         "@contentful/rich-text-types": "^16.0.2",
-        "axios": "^1.6.7",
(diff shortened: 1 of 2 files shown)


mgoudy91 (approved):

seanparmelee:
Heads up that axios was updated again: https://github.com/contentful/contentful.js/pull/2255

PR #2256, opened 2024-06-13
feat: upgrade csm []

2wce:
upgrading csm & fixing tests broken by new behaviour in content api added in this [ticket](https://contentful.atlassian.net/browse/TOL-2035). change will be merged as breaking change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0-determined-by-semantic-release",
       "license": "MIT",
       "dependencies": {
-        "@contentful/content-source-maps": "^0.5.0",
+        "@contentful/content-source-maps": "^0.6.0",
         "@contentful/rich-text-types": "^16.0.2",
         "axios": "~1.7.2",
         "contentful-resolve-response": "^1.8.1",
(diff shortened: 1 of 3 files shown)


t-col (approved):

mgoudy91 (approved):
Looks good to me, though I wonder: if this is a breaking change should we see failing tests to reflect that? Thinking it would be nice to have higher confidence when we deploy that it's working the way we want

2wce:
> Looks good to me, though I wonder: if this is a breaking change should we see failing tests to reflect that? Thinking it would be nice to have higher confidence when we deploy that it's working the way we want

@mgoudy91 I pushed the upgrade on it's own first in this [commit](https://github.com/contentful/contentful.js/pull/2256/commits/5823656d5751834a253fae62c56147cec6875d3e) to show that tests will break then i fixed it in the followup commit

mgoudy91:
> > Looks good to me, though I wonder: if this is a breaking change should we see failing tests to reflect that? Thinking it would be nice to have higher confidence when we deploy that it's working the way we want
> 
> @mgoudy91 I pushed the upgrade on it's own first in this [commit](https://github.com/contentful/contentful.js/pull/2256/commits/5823656d5751834a253fae62c56147cec6875d3e) to show that tests will break then i fixed it in the followup commit

Ah, my mistake. sounds great, thanks for the update @2wce !

PR #2257, opened 2024-06-13
fix: pin axios to 1.6.x and prevent dependabot from overwriting the package manifest

t-col:
[Dependabot forced an update](https://github.com/contentful/contentful.js/pull/2255) to a known incompatible minor version (1.7.x) of axios after we had attempted to pin it at the latest known compatible version (1.6.x). This reverts that change, and also updates the dependabot config to prevent this from happening again.


mgoudy91 (approved):
makes sense, do we need to do this elsewhere?

t-col:
@mgoudy91 probably CMA, yeah. I'll get to that

PR #2283, opened 2024-07-18
fix: typescript for queries on array references

nikmilson:
## Summary

According the docs the queries on references are allowed. While it works for single references, it leads to a ts error on arrays. As it appears to be a typing problem on CDK's side.

https://github.com/contentful/contentful.js/blob/master/TYPESCRIPT.md#limitation

## Description

Add an array of entries to `ReferenceSearchFilters` type to enable the same behavior for array of entries as for single entries.

Example of the wrong behavior:

```typescript
const entries = client.getEntries<
  EntrySkeletonType<
    {
      tags?: EntryFieldTypes.Array<EntryFieldTypes.EntryLink<EntrySkeletonType>>;
    },
    'example'
  >
>({
  content_type: 'example',
  'fields.tags.sys.id[in]': ['foo', 'bar'] // TS Error: Object literal may only specify known properties, and ''fields.tags.sys.id[in]'' does not exist in type 'EntriesQueries<EntrySkeletonType<{ tags?: Array<EntryLink<EntrySkeletonType>> | undefined; }, "example">, undefined>'.ts(2345)
});
const workingEntries = client.getEntries<
  EntrySkeletonType<
    {
      tag?: EntryFieldTypes.EntryLink<EntrySkeletonType>;
    },
    'example'
  >
>({
  content_type: 'example',
  'fields.tag.sys.id': 'foo'
});
``` 

## Motivation and Context

It solves the described TS problem.

## Todos

None.

## Screenshots (if appropriate):
image

### DIFF
--- a/lib/types/query/reference.ts
+++ b/lib/types/query/reference.ts
@@ -1,7 +1,10 @@
 import { EntryFieldTypes } from '../entry'
 import { ConditionalPick } from 'type-fest'
 
-type SupportedTypes = EntryFieldTypes.EntryLink<any> | undefined
+type SupportedTypes =
+  | EntryFieldTypes.Array<EntryFieldTypes.EntryLink<any>>
+  | EntryFieldTypes.EntryLink<any>
+  | undefined
 
 /**
  * Search on references in provided fields


mgoudy91:
@nikmilson Thanks for the contribution! This looks great to me.

One quick request, could you run `npm run prettier` on your fork to fix some enforced linting rules? Other than that this should be good to merge and deploy

PR #2294, opened 2024-07-31
feat: add support for external references

veu:
## Summary

Add support for external references. 

## Description

Since these aren’t resolved they skip the `withoutUnresolvedLinks` behavior and are always kept.

I’ve only updated the `parseEntries` part of the integration tests because it’s enough to cover the simple logic we currently have. The underlying behavior is the same for `getEntries` anyway. If we complicate the logic in the future we can expand the tests as well.

### DIFF
--- a/lib/types/entry.ts
+++ b/lib/types/entry.ts
@@ -42,13 +42,17 @@ export declare namespace EntryFieldTypes {
     type: 'EntryResourceLink'
     entry: EntrySkeleton
   }
+  type ExternalResourceLink = {
+    type: 'ExternalResourceLink'
+  }
   type AssetLink = { type: 'AssetLink' }
   type Array<
     Item extends
       | EntryFieldTypes.Symbol
       | EntryFieldTypes.AssetLink
       | EntryFieldTypes.EntryLink<EntrySkeletonType>
-      | EntryFieldTypes.EntryResourceLink<EntrySkeletonType>,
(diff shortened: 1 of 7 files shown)


mgoudy91 (approved):
Code changes look good, I'll TAL at the build and see if I can see what's up

mgoudy91 (approved):

lib/types/entry.ts
```diff
@@ -243,9 +250,11 @@ export type ResolvedLink<
...
     : Field extends EntryFieldTypes.EntryResourceLink<infer LinkedEntry>
       ? ResolvedEntryResourceLink<Modifiers, Locales, LinkedEntry>
-      : Field extends EntryFieldTypes.AssetLink
-        ? ResolvedAssetLink<Modifiers, Locales>
-        : BaseFieldMap<Field>
+      : Field extends EntryFieldTypes.ExternalResourceLink
+        ? { sys: ResourceLink<string> }
+        : Field extends EntryFieldTypes.AssetLink
+          ? ResolvedAssetLink<Modifiers, Locales>
+          : BaseFieldMap<Field>
```
t-col, line 257: is there no other way to represent this? 😬

t-col (approved):

lib/types/entry.ts
veu: What do you mean by represent?

PR #2299, opened 2024-08-06
BREAKING CHANGE: force release

mgoudy91:
## Todos

-   [x] Implemented feature


marcolink:
@mgoudy91 I think it would be good to add a little description if we merge a PR with "BREAKING CHANGE" (I know this is `next`).

PR #2341, opened 2024-10-01
Fix types not working in ESM mode

### LINKED ISSUE #2340: v11 ships broken types
## Expected Behavior

Consider the following:

```ts
type TypeMyItem = Entry<TypeMyItemSkeleton, 'WITHOUT_UNRESOLVABLE_LINKS'>;
```

This should produce a structure of item like this:

```ts
const item: {
    sys: EntrySys;
    metadata: Metadata;
    sys: {
        contentType: {
            sys: {
                id: "myItem";
            };
        };
    };
    fields: {
      id: string
    }
}
```

## Actual Behavior

Instead, this produces the following structure: 

```ts
const item: {
    sys: EntrySys;
    metadata: Metadata;
    sys: {
        contentType: {
            sys: {
                id: "accordionItem";
            };
        };
    };
    fields: {
        id: {
            [x: string]: string | undefined;
        };
    }
}
```
## Possible Solution

In v11, you shipped support for pure ESM, but forgot to fix module specifiers in types.

Looking at e.g. `contentful/dist/types/types/entry.d.ts`, we can see:

```ts
import { Document as RichTextDocument } from '@contentful/rich-text-types';
import { Asset } from './asset';
import { ContentfulCollection } from './collection';
import { ContentTypeLink, UnresolvedLink } from './link';
import { LocaleCode } from './locale';
import { Metadata } from './metadata';
import { EntrySkeletonType } from './query';
import { EntitySys } from './sys';
import { JsonArray, JsonObject } from 'type-fest';
import { ResourceLink } from './resource-link';
import { ChainModifiers } from './client';
```

Since these imports are all, under ESM conditions, invalid, that makes the types imported from them resolve to `any`. Because of that, in `Entry` type, `ChainModifiers extends Modifiers` condition resolves to `any extends Modifiers` and becomes truthy even if `Modifiers` type parameter is actually provided and valid.

Changing that bit to:

```ts
import { Document as RichTextDocument } from '@contentful/rich-text-types';
import { Asset } from './asset.js';
import { ContentfulCollection } from './collection.js';
import { ContentTypeLink, UnresolvedLink } from './link.js';
import { LocaleCode } from './locale.js';
import { Metadata } from './metadata.js';
import { EntrySkeletonType } from './query.js';
import { EntitySys } from './sys.js';
import { JsonArray, JsonObject } from 'type-fest';
import { ResourceLink } from './resource-link.js';
import { ChainModifiers } from './client.js';
```

immediately fixes the issue.

## Steps to Reproduce

1. Have a repository that actually consumes pure ESM modules ("type": "module" in `package.json`, "module": "nodenext" in `tsconfig.json`…)
2. Try to use the snippet above

## Context

Update to v11 caused hundreds of TypeScript errors on our side caused by invalid types.

Looking at https://arethetypeswrong.github.io/?p=contentful%4011.0.3 confirms my suspicions: the package types are not configured correctly - for both CJS and ESM in fact.

## Environment

-   **Language Version**:  22.8.0
-   **Package Manager Version**: Yarn 4.5.0
-   **Browser Version**: n/a
-   **Operating System**: Darwin MacBook-Air.home 24.0.0 Darwin Kernel Version 24.0.0: Mon Aug 12 20:54:26 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T8112 arm64
-   **Package Version**: 11.0.3

wojtekmaj:
## Summary

Fixes #2340

## Todos

-   [x] Implemented feature

### DIFF
--- a/lib/create-contentful-api.ts
+++ b/lib/create-contentful-api.ts
@@ -5,7 +5,7 @@
 
 import { encodeCPAResponse } from '@contentful/content-source-maps'
 import { AxiosInstance, createRequestConfig, errorHandler } from 'contentful-sdk-core'
-import { CreateClientParams } from './contentful'
+import { CreateClientParams } from './contentful.js'
 import { GetGlobalOptions } from './create-global-options.js'
 import pagedSync from './paged-sync.js'
 import type {
@@ -28,7 +28,7 @@ import type {
 import normalizeSearchParameters from './utils/normalize-search-parameters.js'
 import normalizeSelect from './utils/normalize-select.js'
 import resolveCircular from './utils/resolve-circular.js'
-import getQuerySelectionSet from './utils/query-selection-set'
+import getQuerySelectionSet from './utils/query-selection-set.js'
 import validateTimestamp from './utils/validate-timestamp.js'
 import { ChainOptions, ModifiersFromOptions } from './utils/client-helpers.js'
 import {
@@ -36,8 +36,8 @@ import {
   validateLocaleParam,
   validateRemoveUnresolvedParam,
   validateResolveLinksParam,
-} from './utils/validate-params'
-import validateSearchParameters from './utils/validate-search-parameters'
+} from './utils/validate-params.js'
+import validateSearchParameters from './utils/validate-search-parameters.js'
 
 const ASSET_KEY_MAX_LIFETIME = 48 * 60 * 60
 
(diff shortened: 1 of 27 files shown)


wojtekmaj:
Fellow maintainers, I would also think about ensuring that this problem won't happen again by switching module resolution to nodenext. With this PR we are down from 50 or so errors when switching to just one.

axe312ger:
Good catch, thanks. Thank must have slipped through!

jjolton-contentful (approved):
thanks very much for the contribution @wojtekmaj

PR #2342, opened 2024-10-01
Replace import assertion with import attribute

wojtekmaj:
## Summary

Import assertions are already deprecated. Failing to fix this will cause everyone on Node.js 22 (LTS this month!) unable to push their changes, as build is a part of prepush hook.

## Todos

-   [x] Implemented feature

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,7 +1,7 @@
 import { dirname, resolve } from 'path'
 import { fileURLToPath } from 'url'
 
-import pkg from './package.json' assert { type: 'json' }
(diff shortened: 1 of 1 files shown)


axe312ger:
Do you know if the `with` will also work with older Node versions? We still need to support v18 and v20 for now.

wojtekmaj:
As long as you're on the latest minor, you should be fine.

Import attributes work on Node.js `^18.20.0 || ^19.0.0 || >=20.10.0`.

jjolton-contentful (approved):

Issue #2318, opened 2024-08-27
Bundle Size for the browser version is too big


moonglum:
We are using Contentful on our site for a few pages like the blog. Unfortunately, the chunk size of contentful keeps increasing version by version (a few versions ago, it was 90kb). It is now at **227kb**, and is now the biggest chunk in our entire bundle. 

## Expected Behavior

* I expect a more reasonable bundle size

## Actual Behavior

* It is now by far the biggest dependency we have

## Possible Solution

Allow tree shaking instead of including the entire minified bundle. You always get the entire code, I assume that contains a lot of code for previewing etc. which we don't use:

![image](https://github.com/user-attachments/assets/11533ba9-c644-4044-b328-550d9e208c21)

## Steps to Reproduce

```ts
import {                                                                                                                                                     
      CreateClientParams,                                                                                                                                      
      createClient,                                                                                                                  
} from 'contentful'; 
```

## Context

A bundle size like that is a burden for our customers, who need to download it.

## Environment

-   **Package Version**: 10.15.0
-   **Which API are you using?**: Delivery

moonglum:
I checked our code, as additional information, these are the APIs we are using:

1. We create the client with `createClient` using the `CreateClientParams` type.
2. On that client, we only call `.withoutUnresolvableLinks.getEntries`, `.withoutUnresolvableLinks.getEntry`, `withAllLocales.getEntries` and `.withAllLocales.getEntry`.

paulcrussell:
Hi @moonglum we've noticed something similar on our project and I thought this PR would most likely resolve it.
https://github.com/contentful/contentful.js/pull/2169

@axe312ger is that you're view also?

axe312ger:
Yes my PR should solve it! :)

axe312ger:
Its published as v11! Can you rebundle and share your results @moonglum?

It should now properly tree shake :)

moonglum:
Thanks, Benedikt! Are there any breaking changes? Or do I just keep everything the same?

axe312ger:
@moonglum The API of the package is the same. We "just" changed the wrapping code. You might need to adjust your requires/imports, but besides of that, everything stayed the same.

Changelog coming soon, preview here: https://github.com/contentful/contentful.js/issues/2334#issuecomment-2376213081

moonglum:
Thank you, the hint with ESM is important :+1: I'll try it out and get back to you, @axe312ger

moonglum:
[...]

PR #2358, opened 2024-10-16
add hint about webpack to MIGRATION.md

axe312ger:
Added my webpack hints from https://github.com/contentful/contentful.js/issues/2318#issuecomment-2393549369 to the mirgation readme


mgoudy91 (approved):
huge, thanks!

PR #2363, opened 2024-10-24
feat: give includes basic entry and asset types instead of any

TimBeyer:
## Summary

Gives includes Entry and Asset arrays at least basic skeleton types.
From there they can then conveniently be narrowed down to the desired types.

## Description

Currently the types are `any`, which is difficult to work with.
Casting from the outside is awkward too especially since we have the modifiers that should propagate.
So we should just set the correct types from the start.

## Motivation and Context

I need to do some processing on the includes and don't want to cast them all the time.

### DIFF
--- a/lib/types/entry.ts
+++ b/lib/types/entry.ts
@@ -339,7 +339,7 @@ export type EntryCollection<
 > = ContentfulCollection<Entry<EntrySkeleton, Modifiers, Locales>> & {
   errors?: Array<any>
   includes?: {
-    Entry?: any[]
-    Asset?: any[]
+    Entry?: Entry<EntrySkeletonType, Modifiers, Locales>[]
+    Asset?: Asset<Modifiers, Locales>[]
   }
 }
--- a/test/types/client/getEntries.test-d.ts
+++ b/test/types/client/getEntries.test-d.ts
@@ -18,105 +18,192 @@ type Fields = {
   moreLinks: Entry<LinkedSkeleton>[]
 }
 
-type EntrySkeleton = EntrySkeletonType<Fields, 'content-type-id'>
+// A skeleton with no fields
+type BaseEntrySkeleton = EntrySkeletonType
+
+// Skeleton with some specific fields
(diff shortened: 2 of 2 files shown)


lib/types/entry.ts
```diff
@@ -339,7 +339,7 @@ export type EntryCollection<
 > = ContentfulCollection<Entry<EntrySkeleton, Modifiers, Locales>> & {
   errors?: Array<any>
   includes?: {
-    Entry?: any[]
-    Asset?: any[]
+    Entry?: Entry<EntrySkeletonType, Modifiers, Locales>[]
```
veu, line 342: I think we should not pass down the `Locales` variable here and use `LocaleCode` instead because linked entries might come from other spaces which can have different locales.

TimBeyer: So when we make cross-space requests we basically don't know anymore what locales we're dealing with?
That makes sense, but won't that then also already be broken in the more common case of actually using the link resolution, since there we cannot distinguish?

I'm a bit split between making it easy for the most common use case of not having cross space links, but a set of explicit locales, and ensuring that the types aren't misleading in the case where you use cross space links to spaces with a different set of locales.

I guess since the includes are also not returned in a structure where customers could provide something like a space / locale mapping that means we basically have no way to ever really have locale types that are anything other than `string`.

So I'm wondering, maybe it's better to keep using `Locales` which will probably in much more than 90% of cases be the expected behavior, and which you can at least put all the locales between the different spaces in as a workaround, instead of forcing `includes` to always have untyped locales.

veu: Yeah, you’re right. Even in most cross space cases the locales should be created to match and in cases that are more complicated the locale types probably won’t help much and have to be checked at runtime anyway in which case you would just use `string` and be correct.

veu (approved):

marcolink (approved):
nice 👏 
Maybe we can add a little paragraph to the docs describing the behaviour?

t-col (approved):

PR #2370, opened 2024-11-07
fix: swap out bundlesize for size-limit [EXT-5920]

BobHemphill76:
## Summary

- swap out bundlesize for size-limit due to sec vuln

## Todos

-   [x] Implemented feature


package.json
```diff
@@ -85,10 +85,10 @@
     "prepush": "npm run test:prepush"
   },
   "dependencies": {
-    "@contentful/rich-text-types": "^16.6.1",
```
BobHemphill76, line 88: looks like this just got re-ordered

jjolton-contentful (approved):
nice find!

PR #2412, opened 2024-12-30
feat: add publishedVersion to EntitySys

### LINKED ISSUE #2408: publishedVersion is missing from EntitySys type
Hello, we're using this library to fetch some entries from the delivery api via the `getEntries` method. For these entries, we'd like to examine their published version. I noticed the API & library are returning the `sys.publishedVersion` field (and is mentioned in the docs [here](https://www.contentful.com/developers/docs/references/content-delivery-api/#/introduction/common-resource-attributes)), but it seems to be missing from the TypeScript types in this library.

## Expected Behavior

Given an `Entry`, I'd expect to be able to access the `sys.publishedVersion` field, i.e.

```ts
import * as contentful from 'contentful';

const contentfulClient = contentful.createClient({
  space: 'space',
  accessToken: 'token'
});

contentfulClient
  .getEntry('entryId')
  .then((entry) => console.log(entry.sys.publishedVersion))
```

## Actual Behavior

The code results in a TypeScript error:

```
Property 'publishedVersion' does not exist on type 'EntrySys & { contentType: { sys: { id: string; }; }; }'. ts(2339) 
```

## Possible Solution

I think it's just be a matter of adding `publishedVersion: number;` to the [`EntitySys` type](https://github.com/contentful/contentful.js/blob/7e7a6437e24bcefa1339a33a776e7e3f10c3bd4d/lib/types/sys.ts#L18C18-L27).

## Steps to Reproduce

Copy aforementioned code into a TypeScript project that has `contentful` installed as a dependency or see the TS error on https://codesandbox.io/p/sandbox/serene-voice-h6rdc2

## Context

We're developing a new feature that involves keeping track of which version of a particular entry a user has seen (so we can let users know if/when there's a new version available than the last one they saw) and we were planning on leveraging the `publishedVersion` field for this.

## Environment

-   **Language Version**: node v22.12.0
-   **Package Manager Version**: pnpm v9.15.0
-   **Browser Version**: n/a
-   **Operating System**: Darwin P62P7H941C 23.6.0 Darwin Kernel Version 23.6.0: Thu Sep 12 23:35:29 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T6000 arm64
-   **Package Version**: 11.3.2

seanparmelee:
## Summary

Adds the `publishedVersion` field to the `EntitySys` type.

Closes #2408 

## Motivation and Context

We'd like to reference the published version of a given entry, but the current typing does not include this field despite it existing on the data returned from the `getEntries` call (and also mentioned in the [docs](https://www.contentful.com/developers/docs/references/content-delivery-api/#/introduction/common-resource-attributes)).

## Notes
The doc also mentions the `publishedVersion` field applies to content types, but I did not see that field returned in my testing so I did not add it to the `ContentTypeSys` type.

### DIFF
--- a/lib/types/sys.ts
+++ b/lib/types/sys.ts
@@ -24,4 +24,5 @@ export interface EntitySys extends BaseSys {
   locale?: string
   contentSourceMaps?: CPAContentSourceMaps
   contentSourceMapsLookup?: ContentSourceMapsLookup
+  publishedVersion: number
 }
--- a/test/types/mocks.ts
+++ b/test/types/mocks.ts
@@ -65,6 +65,7 @@ export const entrySys: EntrySys = {
   updatedAt: dateValue,
   id: stringValue,
   createdAt: dateValue,
+  publishedVersion: numberValue,
 }
 
 export const entryBasics = {
@@ -117,6 +118,7 @@ export const assetSys: AssetSys = {
   updatedAt: dateValue,
   id: stringValue,
   createdAt: dateValue,
+  publishedVersion: numberValue,
 }
 
 export const assetBasics = {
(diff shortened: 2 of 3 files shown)


jjolton-contentful (approved):
Thanks for the contribution @seanparmelee

PR #2569, opened 2025-09-26
chore: install content-source-maps from npm instead of github

### LINKED ISSUE #2568: Get 403 when tried to execute npm insall on contentful.js
I want to contribute to contentful.js. So, I checked out the code then execute `npm install` as instructed on https://github.com/contentful/contentful.js/blob/master/CONTRIBUTING.md.

However, npm failed to install `@contentful/content-source-maps` with following error.

> npm error code E403
npm error 403 403 Forbidden - GET https://npm.pkg.github.com/download/@contentful/content-source-maps/0.11.33/2150c1bf0944d0fb1922abce6f86ca216d365ea1 - Permission permission_denied: read_package
npm error 403 In most cases, you or one of your dependencies are requesting
npm error 403 a package version that is forbidden by your security policy, or
npm error 403 on a server you do not have access to.

I've created a GitHub PAT token with read_package and put it in .npmrc already.

elylucasctfl:
The @contentful/content-source-maps dep accidently got installed from our private github packages repository. This fix makes it so its installed from npm instead.

Closes #2568

### DIFF
--- a/test/integration/getConcepts.test.ts
+++ b/test/integration/getConcepts.test.ts
@@ -56,7 +56,7 @@ describe('getConcepts', () => {
     it('filters by Concept Scheme', async () => {
       const response = await client.getConcepts({ conceptScheme: '29lkBedZoW295B4sR7Hwrw' })
 
-      expect(response.items).toHaveLength(2)
+      expect(response.items.length).toBeGreaterThan(0)
     })
   })
 })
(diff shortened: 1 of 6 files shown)


test/integration/getConcepts.test.ts
whitelisab, line 1: Should the changes to this file be removed?

elylucasctfl: the lock file in in sync with it atm. I don't think theres harm in upgrading to the latest patch. dependabot would do it eventually anyways.

whitelisab (approved):

PR #2612, opened 2025-11-28
chore: [MEC-2388] add copilot setup workflow to facilitate using Copilot for Lavamoat setup

stefreak:


stefreak:
Copilot setup only necessary for private repositories

PR #2635, opened 2026-02-12
feat(types): align CreateClientParams with contentful-sdk-core

nib-bturner:
## Summary

A few months ago I added support for `adapter: 'fetch'` and `fetchOptions` to the contentful-sdk-core in https://github.com/contentful/contentful-sdk-core/pull/540. I expected this to flow through automatically - but alas this is not the case. This pull request adds these same type changes to the `CreateClientParams`.

## Motivation and Context

Ideally I'd like to use nextJS's data cache functionality to cache requests to Contentful where they make sense. This change gives use a more powerful set of controls to do this.

## PR Checklist

- [x] I have read the `CONTRIBUTING.md` file
- [x] All commits follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Documentation is updated (if necessary)
- [x] PR doesn't contain any sensitive information
- [x] There are no breaking changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5835,9 +5835,9 @@
       }
     },
     "node_modules/contentful-sdk-core": {
-      "version": "9.4.4",
-      "resolved": "https://registry.npmjs.org/contentful-sdk-core/-/contentful-sdk-core-9.4.4.tgz",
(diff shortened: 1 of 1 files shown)


nib-bturner:
Also noticed a bug in the contentful-sdk-core related to this as well. See: https://github.com/contentful/contentful-sdk-core/pull/566

ethan-ozelius-contentful (dismissed):

PR #2698, opened 2026-06-10
fix(types): distribute Entry over union skeletons in resolved links

TimBeyer:
## Context

When a linked field targets a **union** of skeletons — e.g. one `EntryLink` referencing several content types:

```ts
type SectionSkeleton = HeroSkeleton | CtaSkeleton | FeatureSkeleton | BannerSkeleton | PricingTableSkeleton

interface PageSkeleton extends EntrySkeletonType {
  contentTypeId: 'page'
  fields: {
    sections?: EntryFieldTypes.Array<EntryFieldTypes.EntryLink<SectionSkeleton>>
  }
}
```

…the resolved field collapsed to a single type carrying only the fields common to **every** member:

```ts
const sections: (BaseEntry & {
    sys: { contentType: { sys: { id: "hero" | "cta" | "feature" | "banner" | "pricingTable" } } }
    fields: { internalName: string }   // ← only the field present in ALL members
})[]
```

rather than the expected discriminated union of member entries (`HeroEntry | CtaEntry | …`).

## Root cause

`Entry<>` is a non-distributive mapped type. With a union skeleton, `keyof EntrySkeleton['fields']` resolves over `keyof (A | B)` — the **intersection** of keys — so `fields` drops every member-specific field. (The `id` correctly becomes the union; only `fields` collapses.) The link-resolution path `ResolvedField → ResolvedLink → ResolvedEntryLink → Entry<Union, …>` captures the full union via `infer LinkedEntry`; the collapse happens inside `Entry`.

## Fix

Add `DistributiveEntry<S, M, L>`, which distributes `Entry<A | B>` into `Entry<A> | Entry<B>` (a no-op for a single skeleton), and use it in `ResolvedEntryLink` / `ResolvedEntryResourceLink`.

Distribution is applied **at the link-resolution boundary** (where the linked skeleton is already known) rather than in `Entry` itself. Making `Entry` distributive directly breaks contextual inference in `getEntry`/`getEntries` (e.g. `expectType<Entry<TestEntrySkeleton>>(client.getEntry('id'))` would fall back to the default skeleton), so `Entry` is intentionally kept non-distributive.

> Note: directly instantiating `Entry<Union>` at the top level (not via a link) remains non-distributive by design, for the inference reason above. This PR targets the link-resolution case described in the report.

## Tests

- New `test/types/entry-union-links.test-d.ts` asserts that a resolved `EntryLink<Union>` (single + array, `WITHOUT_UNRESOLVABLE_LINKS`) yields `(Entry<Hero> | Entry<Cta> | undefined)[]`, accepts member-specific fields, and stays discriminable on `sys.contentType.sys.id`. Fails before the fix (2 errors at the collapse sites), passes after.
- `npx tsd` → 0 errors (new + all existing type tests).
- `npm run lint` → 0 errors; `npm run test:unit` → 119 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/types/entry.ts
+++ b/lib/types/entry.ts
@@ -165,6 +165,22 @@ export type BaseFieldMap<Field extends EntryFieldType<EntrySkeletonType>> =
                     : never
 
 /**
+ * Resolves a (possibly union) linked-entry skeleton into the corresponding
+ * `Entry` type, distributing over the union so that `EntryLink<A | B>` becomes
+ * `Entry<A> | Entry<B>` instead of collapsing `fields` to the keys common to
+ * every member. `Entry` itself is intentionally non-distributive to remain
+ * inferrable in `getEntry`/`getEntries`, so distribution is applied here at the
+ * link-resolution boundary where the linked skeleton is already known.
+ * @category Entry
+ * @internal
+ */
+export type DistributiveEntry<
+  EntrySkeleton extends EntrySkeletonType,
+  Modifiers extends ChainModifiers,
+  Locales extends LocaleCode,
+> = EntrySkeleton extends EntrySkeletonType ? Entry<EntrySkeleton, Modifiers, Locales> : never
+
+/**
  * A single resolved link to another entry in the same space
  * If the current client configuration includes `withoutLinkResolution` chain option,
  * the returned type will not resolve linked entities, but keep them as objects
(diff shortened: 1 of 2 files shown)


tylerpina (approved):
🚀

